PR #105, opened 2022-09-23
SSC roles

### LINKED ISSUE #47: CRUD operations for roles in ssc

wtfacoconut:
Initial command implementation for `ssc roles`. `List`, `get`, `create`, and `delete`. Work is still being done on the `update` command.
This PR partially closes #47 .

### DIFF
--- a/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
+++ b/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
@@ -6,10 +6,10 @@ import picocli.CommandLine.Command;
         name = "role",
         aliases = {},
         subcommands = {
-                SSCRoleListCommand.class,
-                SSCRoleGetCommand.class,
                 SSCRoleCreateCommand.class,
-                SSCRoleDeleteCommand.class
+                SSCRoleDeleteCommand.class,
+                SSCRoleGetCommand.class,
+                SSCRoleListCommand.class
         }
 
 )
--- a/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
+++ b/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
@@ -7,7 +7,9 @@ import picocli.CommandLine.Command;
         aliases = {},
         subcommands = {
                 SSCRoleListCommand.class,
-                SSCRoleGetCommand.class
+                SSCRoleGetCommand.class,
+                SSCRoleCreateCommand.class,
+                SSCRoleDeleteCommand.class
         }
 
 )
--- a/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
+++ b/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
@@ -6,7 +6,8 @@ import picocli.CommandLine.Command;
         name = "role",
         aliases = {},
         subcommands = {
-                SSCRoleListCommand.class
+                SSCRoleListCommand.class,
+                SSCRoleGetCommand.class
         }
 
 )
--- a/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
+++ b/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
@@ -1,0 +1,14 @@
+package com.fortify.cli.ssc.role.cli.cmd;
+
+import picocli.CommandLine.Command;
+
+@Command(
+        name = "role",
+        aliases = {},
+        subcommands = {
+                SSCRoleListCommand.class
+        }
+
+)
+public class SSCRoleCommands {
+}
--- a/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
+++ b/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fortify.cli.common.json.JsonHelper;
 import com.fortify.cli.common.output.cli.mixin.IOutputConfigSupplier;
 import com.fortify.cli.common.output.cli.mixin.OutputConfig;
 import com.fortify.cli.common.output.cli.mixin.OutputMixin;
@@ -80,10 +81,7 @@ public class SSCRoleCreateCommand extends AbstractSSCUnirestRunnerCommand implem
         newRole.put("name",name);
         newRole.put("description", description);
         newRole.put("allApplicationRole", allApplicationRole);
-
-        ArrayNode permissionsNode = (new ObjectMapper()).createArrayNode();
-        Arrays.stream(permissionIds).forEach(permissionsNode::add);
-        newRole.set("permissionIds", permissionsNode);
+        newRole.set("permissionIds", JsonHelper.toArrayNode(permissionIds));
 
         outputMixin.write(
                 unirest.post(SSCUrls.ROLES).body(newRole)
--- a/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
+++ b/fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
@@ -1,0 +1,99 @@
+/*******************************************************************************
+ * (c) Copyright 2021 Micro Focus or one of its affiliates
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a 
+ * copy of this software and associated documentation files (the 
+ * "Software"), to deal in the Software without restriction, including without 
+ * limitation the rights to use, copy, modify, merge, publish, distribute, 
+ * sublicense, and/or sell copies of the Software, and to permit persons to 
+ * whom the Software is furnished to do so, subject to the following 
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY 
+ * KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE 
+ * WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR 
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE 
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, 
+ * DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF 
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN 
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS 
+ * IN THE SOFTWARE.
+ ******************************************************************************/
+package com.fortify.cli.ssc.role.cli.cmd;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fortify.cli.common.output.cli.mixin.IOutputConfigSupplier;
+import com.fortify.cli.common.output.cli.mixin.OutputConfig;
+import com.fortify.cli.common.output.cli.mixin.OutputMixin;
+import com.fortify.cli.ssc.rest.SSCUrls;
+import com.fortify.cli.ssc.rest.cli.cmd.AbstractSSCUnirestRunnerCommand;
+import com.fortify.cli.ssc.role.cli.mixin.SSCRoleResolverMixin;
+import com.fortify.cli.ssc.util.SSCOutputConfigHelper;
+import io.micronaut.core.annotation.ReflectiveAccess;
+import kong.unirest.JsonNode;
+import kong.unirest.UnirestInstance;
+import lombok.Getter;
+import lombok.SneakyThrows;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+import picocli.CommandLine.Option;
+
+import java.util.Arrays;
+
+@ReflectiveAccess
+@Command(name = "create")
+public class SSCRoleCreateCommand extends AbstractSSCUnirestRunnerCommand implements IOutputConfigSupplier {
+
+    @Mixin
+    private OutputMixin outputMixin;
+
+/*
+    # NOTE: There are some permissions that have a dependency on another permission.
+    When using the REST API, SSC will not tell you if you have unsatisfied permission dependencies.
+ */
+
+    @Getter
+    @Option(names = {"-n", "--name"}, required = true)
+    private String name;
+
+    @Getter
+    @Option(names = {"-d", "--description"})
+    private String description;
+
+    @Getter
+    @Option(names = "--universal-access", defaultValue = "false")
+    private Boolean allApplicationRole;
+
+    @Getter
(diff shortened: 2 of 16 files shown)


fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
```diff
@@ -0,0 +1,17 @@
+package com.fortify.cli.ssc.role.cli.cmd;
+
+import picocli.CommandLine.Command;
+
+@Command(
+        name = "role",
+        aliases = {},
+        subcommands = {
+                SSCRoleListCommand.class,
```
rsenden, line 9: Can you please put the subcommands in alphabetical order?

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
```diff
@@ -0,0 +1,99 @@
...
+    private String[] permissionIds;
+
+    @SneakyThrows
+    protected Void run(UnirestInstance unirest) {
+        ObjectNode newRole = (new ObjectMapper()).createObjectNode();
+        newRole.put("name",name);
+        newRole.put("description", description);
+        newRole.put("allApplicationRole", allApplicationRole);
+
+        ArrayNode permissionsNode = (new ObjectMapper()).createArrayNode();
```
rsenden, line 84: I think you can use JsonHelper.toArrayNode() here

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleGetCommand.java
```diff
@@ -0,0 +1,56 @@
...
+public class SSCRoleGetCommand extends AbstractSSCGetCommand {
+    @Mixin
+    private SSCRoleResolverMixin.PositionalParameter role;
+
+    protected GetRequest generateRequest(UnirestInstance unirest) {
+        return unirest.get(SSCUrls.ROLE(role.getRoleId(unirest)));
+    }
+
+    @Override
+    protected JsonNode generateOutput(UnirestInstance unirest) {
```
rsenden, line 49: You should only override one of `generateRequest` or `generateOutput`; in this case it looks like you can remove the `generateOutput` method

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleGetCommand.java
```diff
@@ -0,0 +1,56 @@
...
+        return unirest.get(SSCUrls.ROLE(role.getRoleId(unirest)))
+                .asObject(ObjectNode.class).getBody();
+    }
+
+    @Override
+    protected boolean isOutputWrappedInDataObject() { return true; }
```
rsenden, line 55: This is the default, so no need to override

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleListCommand.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fortify.cli.ssc.rest.cli.cmd.AbstractSSCTableOutputCommand;
+import io.micronaut.core.annotation.ReflectiveAccess;
+import kong.unirest.GetRequest;
+import kong.unirest.UnirestInstance;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+@ReflectiveAccess
+@Command(name = "list")
+public class SSCRoleListCommand extends AbstractSSCTableOutputCommand {
```
rsenden, line 39: I think it would make sense to add additional columns `builtIn`, `allApplicationRole` and `description`

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/mixin/SSCRoleResolverMixin.java
```diff
@@ -0,0 +1,81 @@
...
+                        .queryString("q", String.format("name:\"%s\"",roleNameOrId));
+                try{
+                    return request.asObject(ObjectNode.class).getBody().get("data").get(0).get("id").asText();
+                }catch (UnexpectedHttpResponseException | NullPointerException ee){
+                    throw new ValidationException("Unable to find the specified role.");
+                }
+            }
+        }
+        
+        public String getRoleId(UnirestInstance unirestInstance) {
```
rsenden, line 65: This is very unclear; you have two methods (`getRole` and `getRoleId`) that return the same value? Why? Having two methods would only make sense if you have `getRoleDescriptor` and `getRoleId` methods as described in other comments.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCommands.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleGetCommand.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleGetCommand.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleListCommand.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleGetCommand.java
```diff
@@ -0,0 +1,56 @@
...
+import kong.unirest.GetRequest;
+import kong.unirest.UnirestInstance;
+import picocli.CommandLine.Mixin;
+import picocli.CommandLine.Command;
+
+@ReflectiveAccess
+@Command(name = "get")
+public class SSCRoleGetCommand extends AbstractSSCGetCommand {
+    @Mixin
+    private SSCRoleResolverMixin.PositionalParameter role;
```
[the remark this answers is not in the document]
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/mixin/SSCRoleResolverMixin.java
wtfacoconut: That's my mistake. My original plan, as you were suggesting earlier, to create a `RoleDescriptor` class which would have been the return type for `getRole`. I ended up forgetting to properly implement both `getRole` and `getRoleId`.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/mixin/SSCRoleResolverMixin.java
```diff
@@ -0,0 +1,81 @@
...
+import javax.validation.ValidationException;
+
+@ReflectiveAccess
+public class SSCRoleResolverMixin {
+    
+    public static abstract class AbstractSSCRoleMixin {
+        public abstract String getRoleNameOrId();
+
+        @SneakyThrows
+        public String getRole(UnirestInstance unirestInstance){
```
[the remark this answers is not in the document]
wtfacoconut: - Create a RoleDescriptor class that provides the most common role attributes like name and id, and extends JsonNodeHolder for access to other attributes
**DONE**
- Create a helper class with a method getRoleDescriptor(roleNameOrId) that looks up the role in SSC and uses JsonHelper.treeToValue to map this to RoleDescriptor
**DONE**
- Most helper classes load all entities and then use case-insensitive matching on id or name; given that the list of roles usually isn't that large, you could potentially use that approach as well. Alternatively, you could try using a bulk request (see SSCBulkRequestBuilder) to invoke both endpoints with a single request to SSC, to optimize the number of round trips.
**DONE WITH CONDITION:** _I didn't do anything to implement case-insitive searches, but currently searching for roles by `role name` is already already case-insensitive. Only when searching for roles by `role ID` will searches be case-sensitive, which I think is a reasonable constraint._
- You can consider making the fields to be retrieved configurable, such that you can optimally reuse the same getRoleDescriptor method for both resolving the id, and for getting full details. See SSCAppVersionHelper and mixin for example. The getRoleId method would then just retrieve the role id, whereas the get command can retrieve all fields by leaving fields empty/null.
**DONE**

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleListCommand.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Option(names={"--id"}) @SSCFilterQParam @AddAsDefaultColumn
+    private String id;
+
+    @Option(names={"--name"}) @SSCFilterQParam @AddAsDefaultColumn
+    private String name;
+
+    @Option(names={"--builtIn"}) @OutputFilter
```
rsenden, line 48: For consistency, I think we should `--built-in` instead

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleListCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    private Boolean builtIn;
+
+    @Option(names={"--allApplicationRole"}) @OutputFilter
```
rsenden, line 51: Again, probably better to use dashes instead of capitalization. Also, I think it would be better to use the same option name as in the create command, i.e. `--universal-access`

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/mixin/SSCRoleResolverMixin.java
```diff
@@ -0,0 +1,63 @@
...
+    public static abstract class AbstractSSCRoleMixin {
+        public abstract String getRoleNameOrId();
+
+        @SneakyThrows
+        public SSCRoleDescriptor getRole(UnirestInstance unirestInstance){
+            return SSCRoleHelper.getRole(unirestInstance, getRoleNameOrId());
+        }
+        
+        public String getRoleId(UnirestInstance unirestInstance) {
+            return getRole(unirestInstance).getRoleId();
```
rsenden, line 48: Not sure how much data is in a role object and whether the endpoint supports the `fields` parameter, but you may want to change this to `return getRole(unirestInstance, "id").getRoleId();`, and update the `getRole` method to accept a `String... fields` parameter, like you have in the helper class. See `SSCAppVersionResolver` for example.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/helper/SSCRoleHelper.java
```diff
@@ -0,0 +1,72 @@
...
+import kong.unirest.GetRequest;
+import kong.unirest.UnirestInstance;
+
+import javax.validation.ValidationException;
+
+public class SSCRoleHelper {
+    public static final SSCRoleDescriptor getRole(UnirestInstance unirestInstance, String roleNameOrId, String... fields) {
+        GetRequest request = unirestInstance.get(SSCUrls.ROLE(roleNameOrId));
+
+        try{
```
rsenden, line 41: I think role id's are not numeric, right? So you can't tell beforehand whether the user passed a role name or id?
In that case, indeed you'd probably need to call the two endpoints like you're currently doing, but you may want to try combining these in a single request using `SSCBulkRequestBuilder`

rsenden (changes requested):

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleCreateCommand.java
```diff
@@ -0,0 +1,97 @@
...
+    @Mixin
+    private OutputMixin outputMixin;
+
+/*
+    # NOTE: There are some permissions that have a dependency on another permission.
+    When using the REST API, SSC will not tell you if you have unsatisfied permission dependencies.
+ */
+
+    @Getter
+    @Option(names = {"-n", "--name"}, required = true)
```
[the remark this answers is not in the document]
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleDeleteCommand.java
```diff
@@ -0,0 +1,63 @@
...
+import kong.unirest.UnirestInstance;
+import lombok.SneakyThrows;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+
+@ReflectiveAccess
+@Command(name = "delete")
+public class SSCRoleDeleteCommand extends AbstractSSCUnirestRunnerCommand implements IOutputConfigSupplier {
+    @Mixin
+    private SSCRoleResolverMixin.PositionalParameter role;
```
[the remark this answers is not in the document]
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleListCommand.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/cmd/SSCRoleListCommand.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/cli/mixin/SSCRoleResolverMixin.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/helper/SSCRoleHelper.java
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role_permission/cli/cmd/SSCRolePermissionListCommand.java
```diff
@@ -0,0 +1,49 @@
...
+
+@ReflectiveAccess
+@Command(name = "list")
+public class SSCRolePermissionListCommand extends AbstractSSCTableOutputCommand {
+    @Option(names={"--id"}) @OutputFilter @AddAsDefaultColumn
+    private String id;
+
+    @Option(names={"--name"}) @OutputFilter @AddAsDefaultColumn
+    private String name;
+
```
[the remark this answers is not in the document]
wtfacoconut: Done.

fcli-ssc/src/main/java/com/fortify/cli/ssc/role/helper/SSCRoleHelper.java
wtfacoconut: **Done.**
But getting this to work wasn't as straightforward as I hoped it would be. The `SSCBulkRequestBuilder` was pretty easy to use, but there is an odd behavior that I observed when interacting with the SSC bulk API.

When using the bulk request to get information from the `/api/v1/roles` api, if the `q` query parameter is used, I observe that:
-  encoded double quote characters can't be used to surround a string
-  If you use the request builder's `queryParameter()` method, any spaces will be replaced with the `+` character, which causes issues. To avoid any issues, for values supplied to the `name` parameter, a space needs to be encoded as `%20`, not the `+` character. 

So as an example:
- Making a standard GET request to the following url will work, but if I take this exact same url and use it in a bulk request, no data will be retrieved:
  -  `http://192.168.65.213:8080/ssc/api/v1/roles?q=name%3A%22Security+Lead%22&limit=2`

- Instead, I need to ensure that the url is modified to look like the following in order to successfully return data from a bulk API call:
  -  `http://192.168.65.213:8080/ssc/api/v1/roles?q=name:Security%20Lead&limit=2`

PR #118, opened 2022-10-04
Not wrap projectVersionId in quotes & added --limit and --order-by options

wtfacoconut:
For the `ssc event list` command.
The application version ID (project version ID) needs to **not** be surrounded with double quotes when used in the query string.

Also, the list command produces a lot of information. To help with this, I've added the `--limit` and `--order-by` options.


wtfacoconut:
@rsenden. Understood. I've updated the code in the PR to fix the one other issues that is still present, which is that the project version ID shouldn't be wrapped in quotes.

Also, I've captured what you said here in an new issue (#120 ), as it doesn't look like we have an exising one that covers what you're talking about.

PR #385, opened 2023-07-31
fix: fixed filtering nonexisting values not throwing an error, fixes #374

### LINKED ISSUE #374: Filtering on non-existing fields does not throw an error
When trying to filter on non-existing fields there should be an error or at least a warning. Especially when the comparison is part of a multi-condition filter this can easily lead to confusion.

Example:
Single Condition (not too bad):

> .\fcli.exe util sample-data list -q "boolValue==true"
> No data

Multiple Conditions:

> .\fcli.exe util sample-data list -q "boolValue==true || stringValue=='value1'"
>  Id    String value  Long value           Double value            Boolean value  Date value  Date time value            Nested object string value  Nested object boolean value  Nested string aray
>  0     value1        1000                 0.7                     true           2000-01-01  2000-01-01T00:00:00+00:00  nestedObjectValue1          true                         nestedArrayValue3, nestedArrayValue4
>  1     value1        1000                 0.7                     true           2000-01-01  2000-01-01T00:00:00+00:00  nestedObjectValue1          true                         N/A
> ...

In my case I was trying to filter on "booleanValue" using "boolValue" and it took us a while to figure out. Having an error message would be useful.

psmf22:
trying to filter on a non-existing field will now throw an exception, in case of inconsistent source data "has('propertyname')" and "get('propertyname')" can be used as workaround


rsenden (changes requested):
This class is supposed to be an exact copy of the class from spring-integration. We should either:
- Keep this original class, and add our own wrapper/subclass (in a `com.fortify.cli...` package) that delegates to / inherits from this class but with a custom implementation for the `canRead` method
- Move this customized class to `com.fortify.cli...` package, thereby making clear that this is an fcli-specific c;lass instead of the standard class from spring-integration
- Update package.java to indicate that this package contains a patched version of the corresponding classes in spring-integration, and add a comment to this particular line stating that this has been modified from the standard spring-integration implementation

PR #389, opened 2023-08-02
tests updated, slight changes to Fcli.groovy

psmf22:
added test case for non-existing properties, fixed test case for now(period), changed Fcli.groovy to throw an UnexpectedFcliResultException rather than an IllegalStateException


fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/_common/Fcli.groovy
```diff
@@ -142,16 +142,24 @@ public class Fcli {
...
+                    throw new UnexpectedFcliResultException(pfx+"Fcli unexpectedly terminated unsuccessfully\n   "
+                        +stderr.join("\n   "), this)
                 }
             }
             return this
         }
     }
     
+    public static class UnexpectedFcliResultException extends RuntimeException {
+        public FcliResult result;
```
rsenden, line 156: As per Java best practices:
- This field should be `private`, not `public`
- This field should be declared as `final`

You should declare a proper `getResult()` method to allow access to the `result` field value. Notes:
- Based on Groovy concepts, you can still use `e.result` to access the result; Groovy will automatically invoke the corresponding getter for such property references
- Groovy provides annotations for automatically generating such getters and related methods; instead of manually declaring the `getResult` method, I think you can simply use the `@Immutable` annotation on the class (see `FcliResult` class for example)

PR #391, opened 2023-08-02
new tests, feat: sc-sast sensor list

psmf22:


fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCAlertDefinitionSpec.groovy
```diff
@@ -0,0 +1,51 @@
...
+import com.fortify.cli.ftest._common.spec.FcliSession
+import com.fortify.cli.ftest._common.spec.Prefix
+import com.fortify.cli.ftest.ssc._common.SSCAppVersion
+
+import spock.lang.AutoCleanup
+import spock.lang.Requires
+import spock.lang.Shared
+
+@Prefix("ssc.alert-definition") @FcliSession(SSC) 
+class SSCAlertDefinitionSpec extends FcliBaseSpec {
```
rsenden, line 28: This test seems to be assume that alert definitions exist, and even that a specific alert definition with id 84 exists; this may be true for the Fortify Hosted UAT environment, but this test will fail on my local SSC environment for example, as I don't have any alert definitions (at this moment). So, the `list` test should assume either `No Data` or any number of alert definitions. The `get` command should only be executed if the list command returned some data, just taking an arbitrary id from that list, but you'll need to investigate how to best implement something like that (or just disable the `get` test for now)

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCAlertSpec.groovy
```diff
@@ -0,0 +1,42 @@
...
+    
+    def "list"() {
+        def args = "ssc alert list"
+        when:
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                size()>=0
+                if(size()>0) {
+                it[0].replace(' ', '').equals("IdTriggereddateAlertdefinitionnameUsernameAlertmessageApplicationandversionlabel")
```
rsenden, line 38: Same as above for alert-definition; expect either these columns or `No Data`

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCAttributeDefinitionSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+        then:
+            verifyAll(result.stdout) {
+                size()>0
+                it[0].replace(' ', '').equals("IdCategoryGuidNameTypeRequired")
+                it.any { it.contains("DevPhase") }
+            }
+    }
+    
+    def "get.byId"() {
+        def args = "ssc attribute-definition get 5"
```
rsenden, line 43: Can you try to find out whether DevPhase always has id 5? Given that this is a standard attribute, we can assume it exists (as in the `get.byName` test), but potentially it could have a different id depending on SSC version, SSC upgrade history, ...

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCJobSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                size()>0
+                it[0].replace(' ', '').equals("JobnameJobgroupJobclassStateCancellablePriorityCreatetimeStarttimeFinishtime")
+                it.any { it.startsWith(" JOB") }
+            }
+    }
+    
+    def "extractJobName"() {
```
rsenden, line 42: I guess this is meant to extract a job name that can then be used in the next test (as the TODO on the get.byName test will need to be fixed in order for that test to succeed on other SSC instances)? Some feedback:
- This test assumes that the list test is executed first, so this spec needs to have the `@StepWise` annotation
- I don't remember whether you can use SpEL expressions in variable references, but potentially you could simply use `::jobs::get(0),jobName` in the get command, instead of having this intermediate test
- If the above doesn't work, potentially you can do something with the `-o expr=...` option to output only the job name, so you can simply use the contents of `result.out` as input for the `get` command

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCPluginSpec.groovy
```diff
@@ -0,0 +1,44 @@
...
+    
+    def "list"() {
+        def args = "ssc plugin list"
+        when:
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                size()>=0
+                if(size()>0) {
+                    it[0].replace(' ', '').equals("IdPluginidPlugintypePluginnamePluginversionPluginstate")
```
rsenden, line 38: Again, expect either list of plugins or 'No Data', or alternatively, make this a StepWise spec that first installs a plugin (if it's not installed already), then tests the various plugin-related commands like list, enable, disable, ..., and finally removes the plugin again (if it wasn't already installed before we started testing).

Potentially we could develop a dummy 'fcli-test' plugin, such that we can assume that this plugin will only be installed by these tests (so, usually it shouldn't exist unless a previous test run was interrupted, and it's safe to always uninstall it when we're done with the other tests)

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCReportTemplateSpec.groovy
```diff
@@ -0,0 +1,43 @@
...
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                size()>0
+                it[0].replace(' ', '').equals("IdNameTypeTemplatedocidInuse")
+                it.any { it.startsWith(" 1") }
+            }
+    }
+    
+    //TODO add tests for create, get, download, generate-config, delete
```
rsenden, line 42: For get/download/..., either assume standards report bundle has been seeded, or first use the `create` command to install some dummy fcli-test report definition, such that the other commands can assume that this particular definition exists.

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/ssc/SSCUserSpec.groovy
```diff
@@ -0,0 +1,76 @@
...
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                size()>0
+                it[1].equals("id: 1")
+            }
+    }
+    
+    def "get.byMail"() {
+        def args = "ssc user get my_email@fortify.com"
```
rsenden, line 65: This is only true for FH UAT environment; maybe better to again use `@StepWise` and take the email address from the output of the `get` command above.

rsenden (changes requested):

PR #400, opened 2023-08-04
chore: fod tests app|rel|ms

psmf22:


fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/_common/FODAppRel.groovy
```diff
@@ -0,0 +1,84 @@
...
+            "--store $fcliVariableName",
+            {it.expectSuccess(true, "Unable to create application release")})
+        return this
+    }
+    
+    public FODAppRel createMicroservicesApp() {
+        Fcli.run("fod app create $appName:$versionName "+
+            "--description Auto\\ created\\ by\\ test " +
+            "--sdlc-status=Development " +
+            "--release=$appName:$versionName "+
```
rsenden, line 56: I'd suggest having a separate field for the microservice name, instead of using `$appName` as the microservice name. 

That's a potential reason for having static `create*` methods, allowing to have a random microservice name when creating microservice apps, while setting the `microserviceName` field to `null` for non-microservice apps. 

For even more flexibility, maybe we should employ the builder pattern, where each `fcli fod app create` option that may need to be customized (like app-type, and maybe app/ms/rel names) is represented by an instance field and corresponding builder method. For regular Java we have the Lombok `@Builder` annotation, I think Groovy may provide something similar.

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/FoDAppSpec.groovy
```diff
@@ -6,21 +6,145 @@ import com.fortify.cli.ftest._common.Fcli
...
+        then:
+            verifyAll(result.stdout) {
+                it.any { it.contains(webApp.appName) }
+                it.any { it.contains(microservicesApp.appName) }
+                it.any { it.contains(mobileApp.appName) }
+            }
+    }
+    
+    def "get.byId"() {
+        def args = "fod app get ::apps::get(0).applicationId"
```
rsenden, line 72: Why are you using the first record from the list command, rather than the explicitly created web/mobile/microservices app? That would allow you to verify that the `get` command returns the expected `appName` for example.

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/FoDAppSpec.groovy
```diff
@@ -6,21 +6,145 @@ import com.fortify.cli.ftest._common.Fcli
...
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                size()>2
+                it[1].startsWith("applicationId: ")
+            }
+    }
+    
+    def "get.byName"() {
+        def args = "fod app get ::apps::get(0).applicationName"
```
rsenden, line 84: Same as above

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/FoDAppSpec.groovy
```diff
@@ -6,21 +6,145 @@ import com.fortify.cli.ftest._common.Fcli
...
+    
     def "list"() {
         def args = "fod app list"
         when:
             def result = Fcli.run(args)
         then:
             verifyAll(result.stdout) {
-                // TODO Add expectations
+                size()>=1
+                if(size()>1) {
```
rsenden, line 30: For any such if-statements, can you please use proper indentation?

rsenden (changes requested):

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/_common/FODAppRel.groovy
```diff
@@ -0,0 +1,84 @@
...
+ * as constituting an additional warranty. Open Text shall not be 
+ * liable for technical or editorial errors or omissions contained 
+ * herein. The information contained herein is subject to change 
+ * without notice.
+ */
+package com.fortify.cli.ftest.fod._common
+
+import com.fortify.cli.ftest._common.Fcli
+
+public class FODAppRel implements Closeable, AutoCloseable {
```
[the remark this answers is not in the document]
psmf22: because creating an app automatically also creates a release

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/FoDAppSpec.groovy
psmf22: because it doesnt matter at all. the creation of all 3 apps is verified in the verifyCreated step

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/_common/FODAppRel.groovy
psmf22: added an extra field for microservice name, if you want to rework it feel free

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/FoDAppSpec.groovy
rsenden: The id is available through the `FoDAppRel::get` method, right? You're currently using the same variable-based approach to get the id of the first record returned by the list command. As mentioned, getting the details of the app that has just been created allows for additional expectations to be verified, like that the get command returns the expected application name.

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/_common/FODAppRel.groovy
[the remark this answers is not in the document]
rsenden: I guess there's something to say for both. This class operates on the `fcli fod app` entity, and the corresponding command classes are named `FoDApp*Command`, so in that regards it should be named `FoDApp`. Indeed creating an app also automatically creates a release, and as such this class can also be used for testing release-related commands, so in that regards, something like `FoDAppRel` also makes sense. 

However, `FoDAppRel` maybe interpreted as both 'application with release', or 'application release'. This class represents the first, but most people may expect the latter, especially since previously `fcli fod app-rel` was an alias for `fcli fod release`, and the command classes were named `FoDAppRel*Command`. So, when seeing the `FoDAppRel` class name, people may expect that this class represents 'application releases', rather than 'application with implicit release'.

So, if you'd want the class name to represent the concept of 'application with release', I'd suggest using a more descriptive name.

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/_common/FODAppRel.groovy
rsenden: Potentially, fcli could contain a bug where the application name is used instead of the specified microservice name when creating the microservice, hence it's better to have an explicit microservice name. Agree that for now it doesn't make sense to complicate things for having the `microserviceName` set to `null` when creating a non-microservices app.

fcli-core/fcli-common/src/main/java/com/fortify/cli/common/util/FcliDataHelper.java
```diff
@@ -124,7 +124,8 @@ public static final void saveFile(Path relativePath, Object contents, boolean fa
 
     private static void writeFileWithOwnerOnlyPermissions(final Path filePath, final String contents, boolean failOnError) {
         try {
```
rsenden, line 126: Please use try-with-resources block, as otherwise an 'Unreleased Resource' vulnerability will be reported by FoD, i.e., something like:

```
try ( var fos = new FileOutputStream(...); var osw = new OutputStreamWriter(...); var writer = new BufferedWriter(...) ) {

}
```

fcli-core/fcli-common/src/main/java/com/fortify/cli/common/variable/FcliVariableHelper.java
```diff
@@ -115,7 +116,7 @@ public static final Writer getVariableContentsWriter(String variableName, String
         checkVariableName(variableName);
         VariableDescriptor descriptor = createVariableDescriptor(variableName, defaultPropertyName, singular, encrypt);
         saveVariableDescriptor(descriptor);
-        PrintWriter pw = new PrintWriter(Files.newOutputStream(getVariableContentsAbsolutePath(variableName), StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.TRUNCATE_EXISTING));
+        PrintWriter pw = new PrintWriter(new OutputStreamWriter(new FileOutputStream(getVariableContentsAbsolutePath(variableName).toString()), StandardCharsets.UTF_8)) ;
```
rsenden, line 119: Apparently we were not using a try-with-resource block before either, but please change to use try-with-resources.

fcli-other/fcli-functional-test/src/ftest/groovy/com/fortify/cli/ftest/fod/FoDReleaseSpec.groovy
```diff
@@ -0,0 +1,120 @@
...
+    def "verifyCreated"() {
+        def args = "fod release list --store releases"
+        when:
+            def result = Fcli.run(args)
+        then:
+            verifyAll(result.stdout) {
+                it.any { it.contains(app.versionName) }
+            }
+    }
+    /* currently does not work
```
rsenden, line 54: Just wondering, why doesn't this work currently? Fcli bug, ...?

PR #418, opened 2023-08-13
timeout commands

psmf22:


rsenden:
To allow for having different default timeouts for the various product modules (600s for FoD, 60s for all other modules) to match other Fortify integrations (FoDUploader ,FoD Jenkins plugin, ...), and because other low-level connectivity settings like disabling SSL checks are handled by the session login commands, it was decided to have the timeout options also on the session login commands. This has now been implemented in 6a0888676f7043ead19deb69bc080224f4881ad6. As such, this PR will be closed.

PR #431, opened 2023-08-23
Review and improve commands for configuring & starting scans

kadraman:
Made changes to SAST config and start commands as discussed.

Also added "entitlement" commands and moved anything entitlement related to the relevant Helper classes. I have not exposed the commands as they probably don't make sense as end-user commands.


fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartSastCommand.java
```diff
@@ -13,99 +13,63 @@
...
-    
+
     @Mixin private FoDDelimiterMixin delimiterMixin; // Is automatically injected in resolver mixins
     @Mixin private FoDReleaseByQualifiedNameOrIdResolverMixin.PositionalParameter releaseResolver;
-    @Option(names = {"--entitlement-id"})
-    private Integer entitlementId;
-    @Option(names = {"--purchase-entitlement"})
-    private final Boolean purchaseEntitlement = false;
+
+    @Option(names = {"--remediation"})
```
rsenden, line 46: I like the (implementation) simplicity of this approach, but shouldn't we still provide users the option to 'start remediation scan if available'? From a user perspective, wouldn't you actually always want to do a remediation scan if available (assuming this has no technical impact, but does reduce the number of entitlements consumed). If so, shouldn't we just remove any such options, and always use a remediation scan if available?

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartDastCommand.java
```diff
@@ -13,47 +13,48 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Properties;
+
 @Command(name = FoDOutputHelperMixins.StartDast.CMD_NAME)
 public class FoDScanStartDastCommand extends AbstractFoDJsonNodeOutputCommand implements IRecordTransformer, IActionCommandResultSupplier {
     DateTimeFormatter dtf = DateTimeFormatter.ofPattern("MM/dd/yyyy HH:mm");
     @Getter @Mixin private FoDOutputHelperMixins.StartDast outputHelper;
     @Mixin private FoDDelimiterMixin delimiterMixin; // Is automatically injected in resolver mixins
     @Mixin private FoDReleaseByQualifiedNameOrIdResolverMixin.PositionalParameter releaseResolver;
+    @Option(names = {"--assessment-type"}, required = true)
```
rsenden, line 55: Ideally, for consistency with the start-sast command, the start-dast command should assume the scan has already been configured (I think that includes settings like assessment type and entitlement id), and simply copy the existing settings from the scan-setup, rather than providing this as options on the start-dast command. From a user perspective, it should look like this command invokes the 'scan-start-with-defaults` endpoint, even though that doesn't exist (yet) for DAST.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartMobileCommand.java
```diff
@@ -55,7 +50,8 @@ public class FoDScanStartMobileCommand extends AbstractFoDJsonNodeOutputCommand
     @Mixin private FoDReleaseByQualifiedNameOrIdResolverMixin.PositionalParameter releaseResolver;
     private enum MobileAssessmentTypes { Mobile, MobilePlus, Remediation }
     @Option(names = {"--assessment-type"}, required = true)
-    private MobileAssessmentTypes mobileAssessmentType;
+    //private MobileAssessmentTypes mobileAssessmentType;
+    private String mobileAssessmentType;
```
rsenden, line 54: Same comment as for DAST, although we can't implement this for now due to FoD lacking `GET scan-setup` for mobile. As discussed, this command should be hidden/marked as PREVIEW in usage header (and with explanation in usage description)

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan_config/cli/cmd/FoDScanConfigSetupSastCommand.java
```diff
@@ -50,14 +50,13 @@
 @DisableTest(TestType.CMD_DEFAULT_TABLE_OPTIONS_PRESENT)
 public class FoDScanConfigSetupSastCommand extends AbstractFoDJsonNodeOutputCommand implements IRecordTransformer, IActionCommandResultSupplier {
```
rsenden, line 51: I'd like to review this command in more detail, but don't have time. Maybe best to just hide/mark as preview either this command only, or the full `scan-config` command tree?

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/rest/FoDUrls.java
```diff
@@ -41,12 +41,15 @@ public class FoDUrls {
...
+    public static final String STATIC_SCAN_START = STATIC_SCANS + "/start-scan";
+    public static final String STATIC_SCAN_START_WITH_DEFAULTS = STATIC_SCANS + "/start-scan-with-defaults";
+    public static final String STATIC_SCAN_START_ADVANCED = STATIC_SCANS + "/start-scan-advanced";
     public static final String DYNAMIC_SCANS = ApiBase + "/releases/{relId}/dynamic-scans";
     public static final String DYNAMIC_SCANS_IMPORT = DYNAMIC_SCANS + "/import-scan";
     public static final String MOBILE_SCANS = ApiBase + "/releases/{relId}/mobile-scans";
     public static final String MOBILE_SCANS_IMPORT = MOBILE_SCANS + "/import-scan";
     public static final String MOBILE_SCANS_SETUP = MOBILE_SCANS + "/scan-setup";
     public static final String MOBILE_SCANS_START = MOBILE_SCANS + "/start-scan";
+    public static final String ENTITLEMENTS = ApiBase + "/tenant-entitlements";
```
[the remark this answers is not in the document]
kadraman: Yes, you are correct a "regular" user, for example "Developer" role wouldn't be able to access this endpoint. I will keep the commands but remove the use of entitlement validation - which is a shame since this is the only API call we can see whether it is still valid/not expired.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
```diff
@@ -125,6 +125,18 @@ public String toString() {
             }
         }
 
+        public EntitlementFrequencyType toFrequencyType() {
+            switch (this._val) {
```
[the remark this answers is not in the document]
kadraman: OK, I will update to your suggestion, there are quite similar ones in FoDEnums style as it was copied from FortifyUploader plugin. Will change these over time.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/entitlement/cli/cmd/FoDEntitlementCommands.java
```diff
@@ -0,0 +1,26 @@
...
+ * herein. The information contained herein is subject to change
+ * without notice.
+ *******************************************************************************/
+
+package com.fortify.cli.fod.entitlement.cli.cmd;
+
+import com.fortify.cli.common.cli.cmd.AbstractContainerCommand;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "entitlement",
```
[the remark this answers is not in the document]
kadraman: I have "hidden" the commands.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/entitlement/helper/FoDEntitlementHelper.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.time.Instant;
+import java.util.Date;
+import java.util.Optional;
+
+public class FoDEntitlementHelper {
+    @Getter
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    public static final JsonNode renameFields(JsonNode record) {
```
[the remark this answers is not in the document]
kadraman: OK, I have renamed as per suggestion.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartSastCommand.java
```diff
@@ -13,99 +13,63 @@
...
-    private FoDRemediationScanPreferenceTypeMixins.OptionalOption remediationScanType;
-    @Mixin
-    private FoDInProgressScanActionTypeMixins.OptionalOption inProgressScanActionType;
-
     @Mixin private ProgressWriterFactoryMixin progressWriterFactory;
 
-    // TODO: refactor use of FoDEnums and use @JsonValues as per: https://github.com/fortify/fcli/issues/279
     @Override
     public JsonNode getJsonNode(UnirestInstance unirest) {
         try ( var progressWriter = progressWriterFactory.create() ) {
```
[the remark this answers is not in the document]
kadraman: OK, done.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartDastCommand.java
kadraman: Yes it should do, however I'm not sure we can get all the information from the API and might take some time to implement.

I would suggest "hiding" the command for now (2.0.x release?)

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartMobileCommand.java
kadraman: As above

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan_config/cli/cmd/FoDScanConfigSetupSastCommand.java
kadraman: I don't think we should hide this one but I will mark it as PREVIEW

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartSastCommand.java
rsenden: OK, I'm not that familiar with FoD, agree to keep this as an option based on your explanation. However, wouldn't it be more user friendly to provide an option 'use remediation scan if available', rather than plain remediation=true|false?

For single scans, if no remediation scan is available, I'd assume that `--remediation=true` would fail, correct? So, to avoid pipelines from failing, while at the same time minimizing entitlement consumption, for single scans it would make sense for users to specify something like `--remediation=IfAvailable`, wouldn't it?

For subscription scans, I guess `--remediation=IfAvailable` would act the same as `--remediation=No`.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/scan/cli/cmd/FoDScanStartDastCommand.java
rsenden: Yes, hiding this for now would probably be easiest & best

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/rest/FoDUrls.java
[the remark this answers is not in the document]
rsenden: Do we need access to this endpoint to work around some of the existing FoD API limitations/inconsistencies, or would it in general be useful to have the ability for regular users to query available entitlements? Or, conceptually speaking, do we want to have the ability to list/select/otherwise process available entitlements on the client side (in fcli or other integrations), or should this all ideally be handled on the FoD side (like automatically selecting entitlements based on preferences when configuring a scan)?

Can you comment on the ongoing email thread with the FoD team on what API functionality would be required/useful to have for use in fcli (and for what reason), and what user roles should be able to invoke these endpoints?

kadraman:
Removed entitlement command as this is only available for Security Lead user and adds little value - moved the entitlement validation into FoDAssessmentTypeHelper where it is relevant. Simplified SAST scan config and SAST/DAST/Mobile start commands. Marked SAST scan config and DAST/Mobile start commands as PREVIEW.

Issue #432, opened 2023-08-25
SSC: Implement commands and mixin "groupBy" groups


wtfacoconut:
For commands like `fcli ssc issue count`, it has an option called `--by` which allows the user to change which criteria to group issues by. Currently, this option only accepts a guid.
It might be a good idea to create a command for retrieving groupings available for a specified application version as it's cumbersome for the user to manually query the API for this information.

rsenden:
Closing this issue, as the `fcli ssc issue count` command now allows for specifying either group name or id (and the same for filters), and commands have been added to list available groups & filters.

PR #433, opened 2023-08-25
feat: Add `fcli ssc appversion-groupset` command and add group by mixin ...

MikeTheSnowman:
Add `fcli ssc appversion-groupset` command, add group by mixin, and updated `fcli ssc issue count` command to use mixin (see #432)


wtfacoconut:
@rsenden , I've manually tested this and appears to work. 
One thing I'd like your opinion on though is with the name that I chose for this entity. I'm unsure if `groupset` is the best term to use.  To get this info, the REST API you'd use is `issue-selector-set-of-project-version-controller`, and the attribute that we care about being returned from that API call is called `groupBySet`.

I'm also in the process of creating the spock test for `fcli ssc appversion-groupset list` and `fcli ssc appversion-groupset get`, but I need some guidance on how to run the test locally on my system.

wtfacoconut:
@rsenden, sorry for the delay. I've made a update to hide this entity command.

PR #483, opened 2023-12-04
chore: `ssc av create --from` now supports copy of `IssueTemplate,VersionAttributes,UserAccess`

gendry-gh:
chore: `ssc av create --from` now supports copy of `IssueTemplate,VersionAttributes,UserAccess`

By default, the `--copy-from` feature will copy everything from the source app, now including (with the priority): 
- `IssueTemplate` (user input > copy-from > system default) 
- `VersionAttribute` (user input > copy-from > auto-required-attrs) 
- `UserAccess` (user input > copy-from)

### DIFF
--- a/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/access_control/helper/SSCAppVersionUserUpdateBuilder.java
+++ b/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/access_control/helper/SSCAppVersionUserUpdateBuilder.java
@@ -12,9 +12,7 @@
  *******************************************************************************/
 package com.fortify.cli.ssc.access_control.helper;
 
-import java.util.Arrays;
-import java.util.LinkedHashSet;
-import java.util.Set;
+import java.util.*;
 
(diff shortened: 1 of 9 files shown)


mathieures:
If it is possible to merge the fix for the attribute separator right away, it would be amazing!

rsenden:
> If it is possible to merge the fix for the attribute separator right away, it would be amazing!

Unfortunately, it's not that simple. This fix violates existing fcli conventions; all other options that take multiple values use `,` as the (primary) separator. It may be better to change this fix to stick to the original primary separator and use a different separator between the attribute values, like `--attrs ATTR1=ABC|DEF,ATTR2=GHI|XYZ` or similar. We'll need to check the impact of such a change though and whether we need to change this in any other places as well (for example, whether there are any other fcli options that take similar 'lists in list' values).

Andhrei:
Below fix to be moved to another PR:

fix: `ssc create`: the `--attrs` now supports multiple attributes with multiple values

rsenden:
A slightly different implementation is now already available in `develop` branch, so closing this PR.

Issue #449, opened 2023-10-06
SC-SAST: Provide ability to pass custom targs/sargs when starting scan


kadraman:
We have many customers that pass custom `targs/sargs` when starting ScanCentral scans remotely, including:

 - filter files (`-filter`)
 - custom rules (`-rules`)
 - scan precision (`-scan-precision`)/ scan policy (`-scan-policy`)

We need to update `sc-sast start scan` command with additional for `sargs` and `targs`. Although the `scancentral package` command accepts a `targs` options it doesn't appear to accept an `sargs` option and in case, from the documentation:

> Deprecated: Fortify recommends that you use the ‑sargs option directly with the [Start Command](https://www.microfocus.com/documentation/fortify-software-security-center/2310/SC_SAST_Help_23.1.0/Content/CLI.htm?tocpath=_____13#Start).

So we should add these options to our `start` command.

rsenden:
Duplicate of #198, but keeping this one as it contains slightly more information.

rsenden:
There is already PR #487, however:

1. I'm not fully convinced yet on how to best pass scan arguments to fcli; do we have a `--sargs` option taking a string, possibly repeatable `--sargs` option, or have explicit fcli options for each supported scan argument (also see `#2` and `#3`)?
2. PR doesn't cater for local (rule/filterproject template) files to be included in the zip-file submitted to ScanCentral SAST Controller.
3. PR doesn't explicitly check for [supported scan arguments](https://www.microfocus.com/documentation/fortify-software-security-center/2420/SC_SAST_Help_24.2.0/index.htm#cli/sargs-options.htm) before submitting the request to the controller. I think ScanCentral Client performs such a check, however not sure how we could properly do this in fcli as supported arguments may vary across ScanCentral SAST/Fortify SCA versions, and fcli needs to work with all currently supported product versions. We should discuss with @young-s-park.
4. For some reason, the PR contains a lot of unrelated updates to functional tests.

In general, these are the challenges that we face from an fcli perspective:
1. Fcli should work properly with any recent/supported ScanCentral SAST version
2. Ideally, we don’t want to include any explicit logic into fcli as to what scan arguments are supported, as the list of supported arguments can potentially vary across different SC SAST versions
[...]

Issue #198, opened 2022-12-14
Additional sc-sast scan start options


kadraman:
We need to add options for specifying (custom) rules file(s), filter file(s) and project template.
We need to add options for pool selection

rsenden:
Superseded by #449

PR #487, opened 2023-12-08
chore: `sc-sast scan start --sargs` scaRuntimeArguments support added

gendry-gh:
#198 
#449 

The added fcli `--sargs --sca-args` option allows specify scan args along with other arguments such as -filter ([see accepted arguments]( https://www.microfocus.com/documentation/fortify-software-security-center/2320/SC_SAST_Help_23.2.0/index.htm#cli/sargs-options.htm?TocPath=Fortify%2520ScanCentral%2520SAST%2520Command-Line%2520Options%257C_____12))

As of --targs, with current fcli support, there is no need for it.
Currently, we can only start a sast scan using package or mbs. These two options do not support/requires translation arguments (targs are passed to the `scancentral package` cmd, and mbs are already translated)


rsenden:
We'd need to double-check exact SC SAST behavior with regards to interaction between `-sargs` and the `-filter`, `-rules` and `--project-template` options on the `scancentral start` command. These dedicated options are used to include a local filter file, rules file or project template into the scan payload; not sure what happens if you specify something like `-sargs -filter myfilter.txt` without specifying the `-filter` option (would that expect the specified file to be available on the sensor?), or if for example `-filter myfilter.txt` automatically adds the corresponding scan argument.

Andhrei:
[The documentation](https://www.microfocus.com/documentation/fortify-software-security-center/2320/SC_SAST_Help_23.2.0/index.htm#cli/sargs-options.htm?TocPath=Fortify%2520ScanCentral%2520SAST%2520Command-Line%2520Options%257C_____12) lists the supported options :

image

Andhrei:
I just checked, for `-filters filter.txt` scancentral client adds the filter.txt in the zip payload : 
image

And it does so for both cmds : 
- `scancentral start [...] -filters filter.txt`
- `scancentral start [...] -sargs "-filters filter.txt"`

As FCLI doesn't package right now, do we want to do the same ? or (for now) let the user insert the required filters/custom rules files in the zip, along with the right `-sargs`

rsenden:
ScanCentral expects the package.zip file inside another zip file (which is created by fcli on the fly); I guess these extra filer/template/rule files go into the outer zip file, not the zip-file created by `scancentral package` command.

Andhrei:
indeed.

What about adding an `--include-file rule.xml` option to insert a file in the outer zip ?

rsenden:
Superseded by #627, which adds support for including local files in scan payload.

PR #530, opened 2024-05-17
Initial spock test for ssc actions

MikeTheSnowman:
Quick initial spock test for SSC actions. I'll keep building out more tests from here.


rsenden:
Won't be merged as similar tests are now already in place.

Issue #545, opened 2024-05-21
`fcli fod issue ls` include suppressed and fixed issues


MikeTheSnowman:
`fcli fod issue ls` does not have an obvious/easy to have fixed and suppressed issues included in the output. 
For ease of use, there should be two options available to allow for this.

PR #546, opened 2024-05-21
fix: `fcli fod issue list`: Add `--include` option to allow for retrieving `fixed` and/or `suppressed` issues

MikeTheSnowman:
`fcli fod issue ls` didn't have an obvious/easy to have fixed and suppressed issues included in the output. 
This PR adds two options `--include-suppressed` and `--include-fixed`.
I've done some light testing of `--include-suppressed`, including testing with existing options like `--embed`, `-o`, and `-q` and everything appears to be working.

I have not done any testing with `--include-fixed` as I don't have a release with any fixed issues. Though the code for this option is almost exactly the same as `--include-suppressed`, so I expect for this to be okay.

This will address #545.


MikeTheSnowman:
I'll add spock tests later. Sorry, I forgot to include that.

MikeTheSnowman:
The PR has been updated to consolidate the two options to a single `--include` option that accepts parameters (`suppressed` & `fixed`).

Issue #556, opened 2024-06-17
Add `--include` to `ssc av ls`


MikeTheSnowman:
### Enhancement Request

Add an `--include` option to `ssc av ls` to account for the other options (includeInactive, myAssignedIssues, onlyIfHasIssues) exposed by the SSC REST API endpoint:
 
![image](https://github.com/fortify/fcli/assets/5297545/44a97336-7a04-4658-836f-fb35d5bf380d)

rsenden:
This was already implemented in fcli 2.x

PR #557, opened 2024-06-17
Added `--include` to `ssc av list`

MikeTheSnowman:
Created to address #556 .

I've added an `include` option to `ssc av ls`. 
I'm not very confident that I've chosen the best names for the arguments that can be supplied to the `--include` option.


rsenden:
Superseded by ba0c1263e3bafc730037c747c9bcce964dd5cfe6

PR #559, opened 2024-06-26
feat: AWS Security Hub Integration

rohitbaryha1:
added configuration for AWS security hub


rsenden:
@rohitbaryha1, can you please update the following?

- Instead of `-r` / `--report-file` parameter, please use `-f` / `--file` for consistency with other actions and fcli commands.
- Consider adding a `(PREVIEW)` label to action header, and explain in description that future versions may provide direct integration with AWS API.

PR #605, opened 2024-09-20
Attribute Release/Application differentiation and `setup-release` action

### LINKED ISSUE #604: `fcli fod release create`: Improve handling of application attributes in `--attrs` option
### Current Behavior

Given a command like the following:

```bash
fcli fod release create ... --attrs AppAttribute=val1,RelAttr=val2
```

Fcli will update `AppAttribute` on the application, which is likely unexpected and unintended behavior. It is unlikely that anyone is explicitly passing application attributes on release creation, so probably not much of an issue right now.

However, at some point we may want to have the `setup-release` action also support automatically creating the application if it doesn't yet exist. We'd likely re-use the existing `--attrs` action parameter, allowing this parameter to be used to configure both application and release attributes. We'd then simply pass the value of the `--attrs` option to both `fcli fod app create` and `fcli fod release create` commands.

If the app doesn't exist yet, the `fcli fod app create` command would create both app and release with the given attributes. However, if the app exists but version doesn't, we wouldn't want the `fcli fod release create` command to update any existing application attributes.

### Expected Behavior

On the `fcli fod release create` command, we should filter out any application attributes, including only release attributes in the release creation request.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

kadraman:
Includes fix for `--attrs` being able to only update relevant type, e.g. Application/Release (fixes fortify#604)

Working `fcli fod action run setup-release` with attributes and Aviator/OSS options.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
@@ -457,6 +457,7 @@ public class FoDEnums {
     }
 
     public enum AttributeTypes {
+        All(0),
         Application(1),
         Vulnerability(2),
         Microservice(3),
@@ -474,6 +475,8 @@ public class FoDEnums {
 
         public String toString() {
             switch (this._val) {
+                case 0:
+                    return "All";
                 case 1:
                     return "Application";
                 case 2:
@@ -488,6 +491,8 @@ public class FoDEnums {
 
         public static AttributeTypes fromInt(int val) {
             switch (val) {
+                case 0:
+                    return All;
                 case 1:
                     return Application;
                 case 2:
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
@@ -455,4 +455,50 @@ public class FoDEnums {
         @JsonProperty("http,https")
         HTTPandHTTPs;
     }
+
+    public enum AttributeTypes {
+        Application(1),
+        Vulnerability(2),
+        Microservice(3),
+        Release(4);
+
+        private final int _val;
+
+        AttributeTypes(int val) {
+            this._val = val;
+        }
+
+        public int getValue() {
+            return this._val;
+        }
+
+        public String toString() {
+            switch (this._val) {
+                case 1:
+                    return "Application";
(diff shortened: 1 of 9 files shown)


rsenden:
@kadraman Looks like you've also updated the `app create` command to only include app attributes. I think in this case, we'd want to include both app and release attributes, as `app create` will also create a release.

kadraman:
> @kadraman Looks like you've also updated the `app create` command to only include app attributes. I think in this case, we'd want to include both app and release attributes, as `app create` will also create a release.

Yes, I forgot about that - will update it.

kadraman:
Having a quick look at this, as an Application must always be created with a Release do you think it might be better to keep the same behaviour rather than specifying `--app-attrs` and `--release-attrs` (if creating a microservice as well it would also include `--microservice-attrs`)????

I could change the code to cater for this “soley” on `fod app create … --attrs …`
But limit `fod release create..` and `fod app update` and `fod release update` to having `--attrs` only work at the appropriate level.

I could then also change `action setup-release` to just having `--attributes` again.

What do you think?

rsenden:
@kadraman I don't see any new commit yet, but I think it's better to just have a single `--attrs` option. On `app create` this would set both app & release attributes, on `app update` we would ignore release attributes, and on `rel update` and `rel create` we would ignore app attributes. The `setup-release` action can then just have a single `--attrs` option.

rsenden:
I haven't checked any resource bundle updates yet, but please make sure that this behavior is properly documented in the option descriptions, i.e., stating that `app create --attrs` accepts both app & release attributes, whereas for the other commands, `--attrs` ignores app or release attributes.

kadraman:
OK great, I will make this update then and commit.

Kevin
Kevin Lee
Lead Solutions Consultant | UK
OpenText Cybersecurity | Fortify | Debricked
Phone:
(+44) 7429 151807
Website:
www.opentext.com<http://www.opentext.com/>

***@***.***<https://www.microfocus.com/en-us/assets/cyberres/magic-quadrant-for-application-security-testing>

From: Ruud Senden ***@***.***>
Sent: Friday, September 20, 2024 10:48 AM
To: fortify/fcli ***@***.***>
Cc: Kevin A. Lee ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] - Re: [fortify/fcli] Attribute Release/Application differentiation and `setup-release` action (PR #605)


I haven't checked any resource bundle updates yet, but please make sure that this behavior is properly documented in the option descriptions, i.e., stating that app create --attrs accepts both app & release attributes, whereas for the other commands, --attrs ignores app or release attributes.

PR #634, opened 2024-11-07
feat: fcli sc-sast session login: Allow for overriding SC SAST Controller URL (resolves #611)

### LINKED ISSUE #611: `fcli sc-sast session login`: Allow for overriding SC SAST Controller URL
### Enhancement Request

We currently retrieve the controller URL from SSC, which works fine in most cases unless the controller URL points to some internal DNS/IP or Kubernetes cluster name that is not accessible from the system where fcli is being run.

We should provide an option on the `fcli sc-sast session login` command to override the controller URL that's being returned by SSC to allow for correct fcli operation in such scenarios. 

Alternatively, we could raise an SSC enhancement request to allow both internal and external controller URL to be configured, in which case fcli could use the external controller URL. This would be the preferred approach, as it would avoid the need to explicitly specify the controller URL on each `fcli sc-sast session login` invocation.

This issue was originally raised at https://github.com/fortify/github-action/issues/47, but since the GitHub Action utilizes fcli for SC SAST connectivity, this needs to be fixed in fcli first.

SangameshV:
Added support in FCLI to accept an optional argument --ctrl-url which accepts the controller URL and in such cases the user input overrides the controller URL fetched from the SSC configurations.

Example Command:
sc-sast session login --ssc-url="http********************************" --ctrl-url "http**************************" -u********* -p***** -c********!


rsenden:
Superseded by #645

PR #674, opened 2025-01-28
feat: Added command fcli ssc aviator audit

kireetivar:
New `aviator audit` Command in `fcli-ssc` Module  

This PR adds the `fcli ssc aviator audit` command to the `fcli-ssc` module:  

**Command:**  
`fcli.jar ssc aviator audit --av <appVersionNameOrId> -f <optional_file_path> --at <aviator_token> --tn <tenant_name> --au <aviator_grpc_server_url>`  

### Key Features:  
1. **Purpose**: Downloads an FPR from SSC using `--av`, processes it via the new `fcli-aviator` module, and updates the FPR using responses from the Aviator GRPC server before re-uploading to SSC.  
2. **New Module**: `fcli-aviator` handles FPR processing and GRPC communication.  
3. **SSC Integration**: Utilizes existing SSC authentication(session).  

### Workflow:  
1. Download FPR via `appVersionNameOrId`.  
2. Process and update FPR using GRPC server responses.  
3. Upload the updated FPR back to SSC.


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,101 @@
...
+    public JsonNode getJsonNode(UnirestInstance unirest) {
+        try (var progressWriter = progressWriterFactory.create()) {
+            SSCAppVersionDescriptor av = appVersionResolver.getAppVersionDescriptor(unirest);
+            Path tempDir = Files.createTempDirectory("ssc_aviator_");
+
+            File fprFile;
+            if (downloadOptions.getDestination() != null && downloadOptions.getDestination().getFile() != null) {
+                fprFile = downloadOptions.getDestination().getFile();
+            } else {
+                fprFile = tempDir.resolve(String.format("%s_%s.fpr", av.getApplicationName(), av.getVersionName())).toFile();
```
rsenden, line 52: Is it safe to use a fixed file name? Possibly, a user could run multiple 'fcli ssc aviator audit' commands on the same appversion simultaneously. Also, I don't think this is currently being cleaned up after processing has completed?

fcli-core/fcli-aviator/build.gradle
```diff
@@ -0,0 +1,69 @@
+plugins {
+    id 'java'
+    id 'com.google.protobuf' version '0.9.4'
+}
+
+apply from: "${sharedGradleScriptsDir}/fcli-module.gradle"
+
+task zipResources_actions(type: Zip) {
```
rsenden, line 8: This task isn't necessary as fcli-aviator doesn't provide any fcli actions

fcli-core/fcli-aviator/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+}
+
+tasks.withType(JavaCompile) {
+    dependsOn 'generateProto'
+}
+
+dependencies {
+    implementation(project(":fcli-core:fcli-common"))
+
+    implementation("args4j:args4j:2.37")
```
rsenden, line 28: Probably best to add all these dependencies to fcli-bom, and remove the version numbers here? Haven you checked whether we can do without any of these dependencies, or possibly exclude some transitive dependencies? I think you already verified (and added a commit) to make sure that these dependencies operate correctly in GraalVM native binaries? Given the number of dependencies, does this significantly increase fcli jar/executable size?

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,101 @@
...
+import java.nio.file.Path;
+
+
+@Command(name = "audit")
+public class SSCAviatorAuditCommand extends AbstractSSCJsonNodeOutputCommand implements IRecordTransformer, IActionCommandResultSupplier {
+    @Getter @Mixin private OutputHelperMixins.TableNoQuery outputHelper;
+    @Mixin private SSCAppVersionResolverMixin.RequiredOption appVersionResolver;
+    @Mixin private SSCArtifactDownloadOptions downloadOptions;
+    @Mixin private ProgressWriterFactoryMixin progressWriterFactory;
+    @Option(names = {"--tenant-token", "--at"}, required = true) private String tenantToken;
```
rsenden, line 36: As a convention, we try to have at most one long (double-dash) option name, and possibly one short (single-dash) option name. The ‘fcli ssc aviator audit’ command provides multiple long option names for some options:
* For tokens, we usually use -t as short option name, so that might be better than --at, although can be possibly confused with ‘tenant’. Maybe also just have --token as long option name instead of --tenant-token, or maybe --aviator-token is clearer?
* For URLs, we usually provide -u as short option name, we might want to explicitly keep --aviator-token though as --url might be too confusing
* Instead of --tenant-name, I think just --tenant might be better, also consistent with --tenant on ‘fcli fod session login’.
Not sure about the best option names, but something to think about.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,101 @@
...
+            var processedFile = AuditFPR.auditFpr(fprFile, tenantToken, tenantName, aviatorUrl);
+
+            progressWriter.writeProgress("Status: Uploading FPR to SSC");
+            JsonNode uploadResponse = uploadFpr(unirest, processedFile, av);
+            JsonNode dataNode = uploadResponse.get("data");
+            String id = dataNode.has("id") ? dataNode.get("id").asText() : "";
+
+            return av.asObjectNode()
+                    .put("artifactId", id)
+                    .put(IActionCommandResultSupplier.actionFieldName, "UPDATED");
```
rsenden, line 71: Setting the action isn't necessary as this is already handled through getActionCommandResult method; only needed in case you're want to override the static value returned by that method

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,101 @@
...
+            progressWriter.writeProgress("Status: Processing FPR");
+
+            var processedFile = AuditFPR.auditFpr(fprFile, tenantToken, tenantName, aviatorUrl);
+
+            progressWriter.writeProgress("Status: Uploading FPR to SSC");
+            JsonNode uploadResponse = uploadFpr(unirest, processedFile, av);
+            JsonNode dataNode = uploadResponse.get("data");
+            String id = dataNode.has("id") ? dataNode.get("id").asText() : "";
+
+            return av.asObjectNode()
```
rsenden, line 69: Maybe it would be good to return the same/similar data as 'fcli ssc artifact upload', or at least add an @DefaultVariablePropertyName annotation that points to the property containing the artifact id to either this class or SSCAviatorCommands (see SSCArtifactCommands), as this would allow for running something like the following to wait for processing of the audited FPR file to complete:

`fcli ssc aviator audit ... --store my_audit`
`fcli ssc artifact wait-for ::my_audit::`

rsenden (changes requested):

rsenden:
Closing as this is outdated; other PR was raised & merged in `3.x-develop` branch

PR #678, opened 2025-02-07
feat: Added command fcli ssc aviator audit to v3.x-develop

kireetivar:
New `aviator audit` Command in `fcli-ssc` Module  

This PR adds the `fcli ssc aviator audit` command to the `fcli-ssc` module:  

**Command:**  
`fcli.jar ssc aviator audit --av <appVersionNameOrId> -f <optional_file_path> --at <aviator_token> --tn <tenant_name> --au <aviator_grpc_server_url>`  

### Key Features:  
1. **Purpose**: Downloads an FPR from SSC using `--av`, processes it via the new `fcli-aviator` module, and updates the FPR using responses from the Aviator GRPC server before re-uploading to SSC.  
2. **New Module**: `fcli-aviator` handles FPR processing and GRPC communication.  
3. **SSC Integration**: Utilizes existing SSC authentication(session).  

### Workflow:  
1. Download FPR via `appVersionNameOrId`.  
2. Process and update FPR using GRPC server responses.  
3. Upload the updated FPR back to SSC.  

### Additional Changes:

1. Added proper file handling after file download.
2. Removed `task zipResources_actions` as it was unnecessary.
3. Removed some dependencies.
   - After these commits:
     - The `.jar` size increased from 22 MB to 42 MB.
     - The `.exe` size increased from 93 MB to 111 MB.
   - The majority of this size increase comes from the gRPC and proto libraries, which are necessary.
4. Changed variable names in `SSCAviatorAuditCommand.java` to maintain consistency with other commands.
5. Removed the `IActionCommandResultSupplier` key from the returned `JsonNode` object as it was redundant in `SSCAviatorAuditCommand.java`.
6. Added `@DefaultVariablePropertyName("id")` to `SSCAviatorAuditCommand.java` so the artifact ID value can be stored and used in subsequent commands.


fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/config/AviatorLoggingConfigure.java
```diff
@@ -0,0 +1,31 @@
+package com.fortify.cli.aviator.config;
+
+import org.slf4j.LoggerFactory;
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.encoder.PatternLayoutEncoder;
+import ch.qos.logback.core.ConsoleAppender;
+
+public class AviatorLoggingConfigure {
```
rsenden: Fcli already provides logging configuration, so what's this used for exactly? Logging to console may cause issues when used in combination with fcli progress writers.

If this is used to output progress messages to the user, it might be better to use the fcli progress writer instead, for consistency with the remainder of fcli. I can understand that want to keep the aviator module independent of fcli classes, so potentially we could investigate the option of wrapping the progress writer in a logging appender, such that all info/warn messages are indirectly written using the progress writer. Alternatively, we could add an abstraction layer, with the aviator module providing a 'logger' interface, for which an implementation is passed from the fcli command to the AuditFPR class. This fcli-specific implementation could then delegate to the progress writer.

Alternatively, the `fcli ssc aviator audit` command should either not use any progress writer, or the progress writer should be closed before calling `AuditFPR.auditFpr(fprFile, token, tenantName, url);`, and, if needed, re-opened after that call.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/core/AuditFPR.java
```diff
@@ -0,0 +1,83 @@
...
+        AviatorLoggingConfigure.configureLogger();
+
+        logger.info("Starting FPR audit process for file: {}", file.getPath());
+
+        Path extractedPath = ZipUtils.extractZip(file.getPath());
+        logger.debug("Extracted FPR to path: {}", extractedPath);
+
+        try {
+            if (!FPRLoadingUtil.isValidFpr(file.getPath())) {
+                logger.error("Invalid FPR file: {}", file);
```
rsenden: In fcli, we usually don't explicitly log an error when throwing an exception, as the error message is already included in the exception that will be logged (if logging is enabled) and shown to the user.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/core/AuditFPR.java
```diff
@@ -0,0 +1,83 @@
...
+                throw new IOException("Invalid FPR file format.");
```
rsenden: I've recently introduced fcli-specific exception types like `FcliSimpleException`, `FcliBugException` & `FcliTechnicalException`, with `FcliSimpleException` meant for reporting 'simple' errors like these to the user, suppressing most of the stack trace. Again, I can understand that you don't want to use fcli exceptions in Aviator code, but potentially the Aviator module can provide similar exception types, with the `fcli ssc aviator audit` command translating these Aviator-specific exceptions to Fcli-specific exceptions.

fcli-core/fcli-aviator/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+}
+
+apply from: "${sharedGradleScriptsDir}/fcli-module.gradle"
+
+tasks.withType(JavaCompile) {
+    dependsOn 'generateProto'
+}
+
+dependencies {
+    implementation(project(":fcli-core:fcli-common"))
```
rsenden, line 13: Are you actually using any classes from fcli-common? Or is this just to include some dependencies like logging framework? If the latter, maybe it's better to explicitly list those dependencies here (based on the versions provided in fcli-bom), instead of depending on fcli-common, if you want to make sure that the Aviator module doesn't use any fcli-specific code. (my assumption is that eventually you may want to re-use the Aviator code in SSC for example, and therefore want to avoid any dependencies on fcli-specific code)

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Comparator;
+
+@Command(name = "audit")
+@DefaultVariablePropertyName("id")
+public class SSCAviatorAuditCommand extends AbstractSSCJsonNodeOutputCommand implements IRecordTransformer, IActionCommandResultSupplier {
+    @Getter @Mixin private OutputHelperMixins.TableNoQuery outputHelper;
+    @Mixin private SSCAppVersionResolverMixin.RequiredOption appVersionResolver;
+    @Mixin private ProgressWriterFactoryMixin progressWriterFactory;
+    @Option(names = {"-t", "--token"}, required = true) private String token;
+    @Option(names = {"--tenant"}, required = true) private String tenantName;
```
rsenden: As mentioned in email; would it be possible/feasible to include/encode the tenant inside the token, such that users only need to pass the -t/--token option? Could be as simple as `<tenantName>:<token>`, but we should then provide the token in this format to users (i.e., users shouldn't need to manually combine `<tenantName>` and `<token>`, but should just be able to copy/paste the combined token that we provide to them). Taking this a step further, potentially we could also include/encode the URL inside the token, such that users only need to copy/paste a single 'combined token', from which this audit command can parse the required URL, tenant name and actual token.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_main/cli/cmd/SSCCommands.java
```diff
@@ -51,6 +52,7 @@
                 SSCAppVersionCommands.class,
                 SSCArtifactCommands.class,
                 SSCAttributeCommands.class,
+                SSCAviatorCommands.class,
```
rsenden, line 55: As mentioned in email; this `aviator` container command contains only a single `audit` leaf command. That's fine if we plan on adding other Aviator commands, but if not, maybe we should look into moving the 'audit' command into one of the existing entities, like `fcli ssc appversion (aviator-)audit` (I think we're processing the application file, so having this in `appversion` probably makes more sense than having this in `artifact`)

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+            fprFile = File.createTempFile(String.format("aviator_%s_%s_", av.getApplicationName(), av.getVersionName()), ".fpr", tempDir.toFile());
+
+            SSCFileTransferHelper.download(
+                    unirest,
+                    SSCUrls.DOWNLOAD_CURRENT_FPR(av.getVersionId(), true),
+                    fprFile,
+                    SSCFileTransferHelper.ISSCAddDownloadTokenFunction.ROUTEPARAM_DOWNLOADTOKEN);
+
+            progressWriter.writeProgress("Status: Processing FPR");
+            processedFile = AuditFPR.auditFpr(fprFile, token, tenantName, url);
```
rsenden: See earlier feedback on the use of a console logger inside the Aviator module; this might interfere with the progress writer. So, either:
- Console logging needs to be adjusted in the Aviator module to use fcli progresswriter instead
- Or the progress writer should be closed before calling `AuditFPR::auditFpr`, and re-opened afterwards
- Or this command should refrain from using a progress writer altogether

Also minor feedback regarding capitalization; I'd try to keep this consistent across class names and method names, i.e., either `AuditFpr.auditFpr` or `AuditFPR.auditFPR`.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+@Command(name = "audit")
+@DefaultVariablePropertyName("id")
+public class SSCAviatorAuditCommand extends AbstractSSCJsonNodeOutputCommand implements IRecordTransformer, IActionCommandResultSupplier {
+    @Getter @Mixin private OutputHelperMixins.TableNoQuery outputHelper;
+    @Mixin private SSCAppVersionResolverMixin.RequiredOption appVersionResolver;
+    @Mixin private ProgressWriterFactoryMixin progressWriterFactory;
+    @Option(names = {"-t", "--token"}, required = true) private String token;
+    @Option(names = {"--tenant"}, required = true) private String tenantName;
+    @Option(names = {"-u", "--url"}, required = true) private String url;
+    private static final Logger logger = LoggerFactory.getLogger(SSCAviatorAuditCommand.class);
```
rsenden: By convention, as this is a static final field, we usually name this field `LOG` instead of `logger`.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    @SneakyThrows
+    public JsonNode getJsonNode(UnirestInstance unirest) {
+        Path tempDir = null;
```
rsenden: Depending on the outcome of some of the discussions below, might be better to remove some/all of these variables being initialized to `null`, and instead use `var <variable> = <value>` at the location where these variables are actually initialized, for example just having `var processedFile = AuditFPR.auditFpr(...)`, instead of first declaring `processedFile` here, and initializing it later.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+            JsonNode dataNode = uploadResponse.get("data");
+            String id = dataNode.has("id") ? dataNode.get("id").asText() : "";
+
+            return av.asObjectNode()
+                    .put("artifactId", id);
+        } catch (Exception e) {
+            logger.error("Error during Aviator audit process", e);
+            throw e;
+        } finally {
+            cleanupResources(tempDir);
```
rsenden: Missed this when writing my email and my earlier comment above about declaring `tempDir` outside of try-with-resources block; still might be better to (also) use File.deleteOnExit, as I think the file will then also be deleted if for example Ctrl-C is pressed. Not sure whether that also works for directories though.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    @SneakyThrows
+    public JsonNode getJsonNode(UnirestInstance unirest) {
+        Path tempDir = null;
+        File fprFile = null;
+        File processedFile = null;
+
+        try (var progressWriter = progressWriterFactory.create()) {
+            SSCAppVersionDescriptor av = appVersionResolver.getAppVersionDescriptor(unirest);
+            tempDir = Files.createTempDirectory("ssc_aviator_");
```
rsenden: Seems like you're only storing a single file in this temp dir, so why even create a temp dir? Easier to just create a temp file with `deleteOnExit`, simplifying the implementation for this command, especially given the fairly complex recursive temp dir cleanup code in the `cleanupResources` method.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    public String getActionCommandResult() {
+        return "UPDATED";
+    }
+
+    @Override
+    public boolean isSingular() {
+        return true;
+    }
+
+    private void cleanupResources(Path tempDir) {
```
rsenden: If we really want to have the temp dir, there's already a `FileUtils::deleteRecursive` method in fcli-common that basically does the same as this method (although it uses `@SneakyThrows`, so you might want to keep the exception handling to just log a warning instead of having an exception thrown)

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
```diff
@@ -0,0 +1,123 @@
...
+import picocli.CommandLine.Command;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Comparator;
+
+@Command(name = "audit")
+@DefaultVariablePropertyName("id")
```
rsenden: Assuming that users may want to invoke `fcli ssc artifact wait-for ::fcliVarName::` to wait for SSC to finish processing the audited artifact (with this audit command being invoked with `--store fcliVarName`), we'd want to change the default variable property name to `artifactId`.

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/i18n/SSCMessages.properties
```diff
@@ -625,6 +633,7 @@ fcli.ssc.appversion.copy-state.output.table.options = previousProjectVersionId,p
 fcli.ssc.artifact.output.table.options = id,scanTypes,lastScanDate,uploadDate,status
 fcli.ssc.attribute.output.table.options = id,category,guid,name,valueString
 fcli.ssc.attribute.definition.output.table.options = id,category,guid,name,type,required
+fcli.ssc.aviator.output.table.options = id,application.name,name,active,issueTemplateName,createdBy
```
rsenden: Do we want to include `artifactId`? Maybe we don't need some of the other properties like `active`, `issueTemplateName`, `createdBy`, as these are not relevant for the audit operation. Also, plain `id` might be confusing as to whether this refers to the appversion id or the artifact id, so maybe we should have a column header that makes it more clear that this is the application version id (or just remove `id` here?)

rsenden (changes requested):

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/core/AuditFPR.java
kireetivar: Fixed this 👍

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/core/AuditFPR.java
kireetivar: Implemented this! 👍 Added appropriate exception handling based on the suggested approach.

fcli-core/fcli-aviator/build.gradle
kireetivar: I agree that fcli-common isn't currently used in this command. However, in the next PR, we'll introduce Aviator-specific commands (e.g., `fcli aviator project list`), where I'll be using the crypto package for message signing. This dependency is kept for that reason.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Agreed! Thanks for idea 😄. I've implemented this change for `tenantName`, but including the `URL` in the token won't be feasible due to various constraints.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Sorry missed this, changed this in new commit.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Agree! Made this change in new commit

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Thanks for suggestion using `File.deleteOnExit` as i switched back to using file.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Agreed! I initially thought using `tempDir` would be better for file isolation, but given the current setup, it's unnecessary. A file will suffice. Thanks for the suggestion! 👍

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Removed this as it's no longer needed; switched back to using a file.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: Good find,  missing this detail ! Changed the default variable property name to artifactId in new commit.

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/i18n/SSCMessages.properties
kireetivar: removed 'id' and other options and added 'artifactId'.

kireetivar:
Sorry ! my bad, I didn’t check your email before replying here 😅 I’m currently looking into solutions for the logging issue and will update the PR once it’s resolved.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/config/AviatorLoggingConfigure.java
kireetivar: Thanks for the feedback! We've removed AviatorLoggingConfigure.java to eliminate console conflicts with fcli's `ProgressWriter`. To address the progress message output, I introduced the `IAviatorLogger` abstraction in `fcli-aviator` and implemented it as `AviatorLoggerImpl`, which uses SLF4J for both console and file logging. Please let me know if any further changes are required.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/aviator/cli/cmd/SSCAviatorAuditCommand.java
kireetivar: The newer implementation appears to resolve this clash. Please take a look and let me know your thoughts. Thanks!

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/i18n/SSCMessages.properties
kireetivar: I’ve added the `id` back in, moved `artifactId` to the end, and removed the `active` field.

PR #693, opened 2025-03-06
FoD: Implement `fcli fod issue update` command (resolves #669)

### LINKED ISSUE #669: FoD: Add support for bulk update of issues/vulnerabilities
### Enhancement Request

A customer has requested the ability to bulk update issues, in particular to set the "auditorStatus" to "Suppressed". This is required because they distribute a list of the issues to a vendor via a CSV/XLS they are marked up by the vendor and they have to manually apply all these changes in the FoD UI.

This should be possible using the `/api/v3/releases/{releaseId}/vulnerabilities/bulk-edit` API endpoint - although they would really want to do it at the application level or the same issue that exists in all releases - like you can do in the UI.

Maybe this could be an action but we should probably include individual commands as well, e.g.

`fcli fod issue bulk-update --release "1234" --issueIds "123,1234" ...`

kadraman:
This PR includes the `fcli fod issue update` command which can be used to bulk audit issues/vulnerabilities. There are no FoD API commands to update individual issues only the bulk update one (which can be used to update an individual one anyway).

I have had a couple of customers asking how they update the audit information for multiple issues and this will command will serve this purpose. It could also be called from any new actions as well.

I have also update the `fcli fod issue ls` to show the `vulnId` field which this command requires/operates on.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
@@ -506,4 +506,104 @@ public class FoDEnums {
         }
     }
 
+    public enum VulnerabilitySeverityType {
+        // use same integer values as FoD uses internally
+        Low(1),
+        Medium(2),
+        High(3),
+        Critical(4),
+        Info(-2),
+        BestPractice(-1);
+
+        private final int _val;
+
+        VulnerabilitySeverityType(int val) {
+            this._val = val;
+        }
+
+        public int getValue() {
+            return this._val;
+        }
+
+        public String toString() {
+            switch (this._val) {
+                case 1:
+                    return "Low";
+                case 2:
+                    return "Medium";
+                case 3:
+                    return "High";
+                case 4:
+                    return "Critical";
+                case -1:
+                    return "Best Practice";
+                case -2:
+                    return "Info";
+                default:
+                    return "Low";
+            }
+        }
+
+        public static VulnerabilitySeverityType fromInt(int val) {
+            switch (val) {
+                case 1:
+                    return Low;
+                case 2:
+                    return Medium;
+                case 3:
+                    return High;
+                case 4:
+                    return Critical;    
+                case -1:
+                    return BestPractice;
+                case -2:
+                    return Info;
+                default:
+                    return Info;
+            }
+        }
+    }
+
+    public enum DeveloperStatusType {
+        // no internal integer id representation
+        Open("Open"),
+        InRemediation("In Remediation"),
+        Remediated("Remediated"),
+        WillNotFix("Will Not Fix"),
+        ThirdPartyComponent("Third Party Component");
+
+        public final String value;
+
+        DeveloperStatusType(String value) {
+            this.value = value;
+        }
+
+        public String getValue() {
+            return this.value;
+        }
+    }
+
+    public enum AuditorStatusType {
+        // no internal integer id representation
+        PendingReview("Pending Review"),
+        RemediationRequired("Remediation Required"),
+        RemediationDeferred("Remediation Deferred"),
+        RiskMitigated("Risk Mitigated"),
+        // the following are used by Aviator and should not be set by a user
+        //Suspicious("Suspicious"),
+        //ProposedNotAnIssue("Proposed Not an Issue"),
+        RiskAccepted("Risk Accepted"),
+        NotAnIssue("Not an Issues");
+
+        public final String value;
+
+        AuditorStatusType(String value) {
+            this.value = value;
+        }
+
+        public String getValue() {
+            return this.value;
+        }
+    }
+
 }
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueUpdateCommand.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueUpdateCommand.java
@@ -1,0 +1,99 @@
+/*******************************************************************************
+ * Copyright 2021, 2025 Open Text.
+ *
+ * The only warranties for products and services of Open Text 
+ * and its affiliates and licensors ("Open Text") are as may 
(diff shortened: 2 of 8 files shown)


fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
```diff
@@ -506,4 +506,104 @@ public static AttributeTypes fromInt(int val) {
...
+
+        VulnerabilitySeverityType(int val) {
+            this._val = val;
+        }
+
+        public int getValue() {
+            return this._val;
+        }
+
+        public String toString() {
```
rsenden, line 528: I know that other enums in `FoDEnums` use similar switch-statements, but I'm not a big fan of this approach (in both `toString` and `from*` methods) as it's error-prone, for example if someone adds a new enum value but forgets to update the `switch` statements.

For `toString`, we can often derive the string value from the enum entry name. For example, in this case, we could insert a space before any uppercase letters (`BestPractice`->`Best Practice`), or probably even easier, use underscores in the enum entries to represent spaces (i.e., `Best_Practice`), and then have `toString` return `name().replace('_', ' ')`. If such an automated mapping is not possible/viable, the string representation should be passed as constructor parameter, such that `toString` can simply return the string representation that was passed through the constructor for each enum entry.

For `from*` or similar methods, it would be better to use one of the approaches listed here: https://softwareengineering.stackexchange.com/questions/445637/getting-an-enum-value-by-a-unique-property-in-java. Personally, I'd prefer using `Stream.of(values()).filter(...).findFirst().orElse...`, but simple for-loop is also fine.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
```diff
@@ -506,4 +506,104 @@ public static AttributeTypes fromInt(int val) {
         }
     }
 
+    public enum VulnerabilitySeverityType {
```
rsenden, line 509: I know similar field & method names are used in other enums in `FoDEnums`, but can we improve these to make it more clear what values are sent to/received from FoD, for example using field name `fodValue` and methods `getFoDValue` and `fromFoDValue`?

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
```diff
@@ -506,4 +506,104 @@ public static AttributeTypes fromInt(int val) {
...
+        Low(1),
+        Medium(2),
+        High(3),
+        Critical(4),
+        Info(-2),
+        BestPractice(-1);
+
+        private final int _val;
+
+        VulnerabilitySeverityType(int val) {
```
rsenden, line 520: It's shorter to remove this constructor, and instead add `@RequiredArgsConstructor` Lombok annotation on the enum class

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
```diff
@@ -506,4 +506,104 @@ public static AttributeTypes fromInt(int val) {
...
+            this._val = val;
+        }
+
+        public int getValue() {
```
rsenden, line 524: Preferably use Lombok `@Getter` on the field instead, once field is renamed to more meaningful name.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueUpdateCommand.java
```diff
@@ -0,0 +1,99 @@
...
+    @Option(names = {"--auditor-status"}, required = false)
+    protected AuditorStatusType auditorStatus;
+    @Option(names = {"--severity"}, required = false)
+    protected VulnerabilitySeverityType severity;
+    @Option(names = {"--comment"}, required = false)
+    protected String comment;
+    @Option(names = {"--ids"}, required = true, split=",")
+    protected ArrayList<String> vulnIds;
+
+    private long errorCount = 0;
```
rsenden, line 61: We should try to avoid having instance fields to avoid race conditions. In this case, `getActionCommandResult` should always return `ISSUES_UPDATED`, and `getJsonNode` should override this using `put(IActionCommandResultSupplier.actionFieldName, "ISSUES_UPDATED_WITH_ERRORS")`

rsenden (changes requested):
For `FoDEnums`, it would be great if all enums can be updated as per these review comments and the existing `TODO` comments in that class, but I don't mind leaving this for later and merging the PR without those enum changes, but can you add least add the information from the review comments as additional `TODO` comments in `FoDEnums` so we don't forget?

For `FoDEnums`, please note that I added a separate review comment about field/method names (the comment was initially added in the wrong place, so I deleted it and tried to re-add, but couldn't figure out how to re-add the comment to the existing review).

The `errorCount` instance field should be fixed before PR is merged.

More in general; FoD recently introduced support for custom issue attributes, even though I think this is a UI-only feature for now; no API to set custom attribute values for example. However, if such an API is added, how would that affect this command? Would we likely be able to integrate custom attribute support in this command, would we need to add another fcli command, ...?

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/util/FoDEnums.java
```diff
@@ -506,4 +506,104 @@ public static AttributeTypes fromInt(int val) {
...
+    public enum VulnerabilitySeverityType {
+        // use same integer values as FoD uses internally
+        Low(1),
+        Medium(2),
+        High(3),
+        Critical(4),
+        Info(-2),
+        BestPractice(-1);
+
+        private final int _val;
```
rsenden, line 518: I know similar field & method names are used in other enums in FoDEnums, but can we improve these to make it more clear what values are sent to/received from FoD, for example using field name fodValue and methods getFoDValue and fromFoDValue?

Issue #740, opened 2025-05-16
`fcli ssc session login`: Allow for bypassing SC-SAST/SC-DAST connection test


rsenden:
### Enhancement Request

With the new combined SSC/SC-SAST/SC-DAST session management in fcli 3.x, the `fcli ssc session login` command will fail if SC-SAST and/or SC-DAST has been configured in SSC but either SC-SAST or SC-DAST is (temporarily) unavailable. This effectively blocks users from using any `fcli ssc` commands if there's an issue with SC-SAST/SC-DAST (if enabled in SSC), even if the user isn't interested in running any `fcli sc-sast` or `fcli sc-dast` commands.

We should allow fcli users to interact with SSC even if there's an issue with SC-SAST or SC-DAST, or for example allowing to interact with SC-SAST if there's an issue with SC-DAST only. 

We'd need to investigate the best approach for handling this situation, including the appropriate 'default' behavior, while keeping backward compatibility. Some potential approaches that come to mind:

- Provide command-line options to simply disable SC-SAST/SC-DAST functionality, either:
    - Explicitly, like `--disable-sc-dast` & `--disable-sc-sast`
    - Implicitly; SC-SAST is already automatically disabled if no `--client-auth-token` is specified, and maybe (need to verify) if `--sc-sast-url` is set to a blank value. SC-DAST doesn't have an `--client-auth-token` equivalent, but maybe we could introduce `--sc-dast-url` that can be set to an empty value to disable SC-DAST (with the added advantage that the URL as configured in SSC can be overridden in fcli)
- Just produce a warning message and zero exit code instead of exception and non-zero exit code if SC-SAST/SC-DAST connection validation fails, such that the `fcli ssc session login` command will succeed but any subsequent `fcli sc-sast` and `fcli sc-dast` commands will then fail with an appropriate exception. 
    - Main drawback is that this is inconsistent with current fcli behavior. For example, if a pipeline runs both SC-SAST & SC-DAST scans, the pipeline will currently fail upon `fcli ssc session login` and thus not run any scans. With the new behavior, the pipeline may now be able to successfully start the SC-SAST scan but then fail to start the SC-DAST scan if SC-DAST is temporarily down.
- Provide command-line options to specify the action to be taken on connection validation failures, like `--on-sc-sast-error=ignore|warn|fail`
    - Should the default value be `fail` (consistent with current fcli behavior), or `warn` (inconsistent with current fcli behavior as described above)

rsenden:
[...]

PR #761, opened 2025-06-20
feat: `fcli ssc session login`: Allow for bypassing SC-SAST/SC-DAST connection test(#740)

jmadhur87:
Add --disable option to bypass the SC-SAST and SC-DAST session while creating SSC session.
image

### DIFF
--- a/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/cmd/SSCSessionLoginCommand.java
+++ b/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/cmd/SSCSessionLoginCommand.java
@@ -64,20 +64,30 @@ public class SSCSessionLoginCommand extends AbstractSessionLoginCommand<SSCAndSc
         try ( var unirest = GenericUnirestFactory.createUnirestInstance() ) {
             testAuthenticatedSCSastConnection(unirest, sessionData);
         }
-        catch(UnirestException  expt) {
-        	logoutBeforeNewLogin(sessionName, sessionData);
-        	getSessionHelper().destroy(sessionName);
-        	String scSastUrl = sessionData.getScSastUrlConfig().getUrl();
-        	throw new FcliSimpleException(String.format("SC-SAST is temporarily unavailable. You can still connect to SSC by using the --disable=sc-sast argument.\nSSC URL: %s\nSC-SAST URL: %s" , sscUrl,scSastUrl), expt.getMessage());
-        }
-        try ( var unirest = GenericUnirestFactory.createUnirestInstance() ) {
-            testAuthenticatedSCDastConnection(unirest, sessionData);
-        }
-        catch(UnirestException  expt) {
-        	logoutBeforeNewLogin(sessionName, sessionData);
-        	getSessionHelper().destroy(sessionName);
-        	String scDastUrl = sessionData.getScDastUrlConfig().getUrl();
-        	throw new FcliSimpleException(String.format("SC-DAST is temporarily unavailable. You can still connect to SSC by using the --disable=sc-dast argument.\nSSC URL: %s\nSC-DAST URL: %s", sscUrl,scDastUrl), expt.getMessage());
+		catch (UnirestException e) {
+			logoutBeforeNewLogin(sessionName, sessionData);
+			getSessionHelper().destroy(sessionName);
+			String scSastUrlConfiguredInSSC = sessionData.getScSastUrlConfig().getUrl();
+			String userGivenScSastUrl = sessionLoginOptions.getSscAndScanCentralUrlConfigOptions().getScSastControllerUrl();
+			if (userGivenScSastUrl != null) {
+				throw new FcliSimpleException(
+						String.format("Unable to connect to the given SC-SAST URL.\nSSC URL: %s\nSC-SAST URL: %s",
+								sscUrl, userGivenScSastUrl),
+						e);
+			} else {
+				throw new FcliSimpleException(String.format(
+						"Unable to connect to SC-SAST URL as configured in SSC; please contact your SSC administrator, or use the --disable option to disable SC-SAST functionality for this session.\nSSC URL: %s\nSC-SAST URL: %s",
+						sscUrl, scSastUrlConfiguredInSSC), e);
+			}
+
+		}
+		try (var unirest = GenericUnirestFactory.createUnirestInstance()) {
+			testAuthenticatedSCDastConnection(unirest, sessionData);
+		} catch (UnirestException e) {
+			logoutBeforeNewLogin(sessionName, sessionData);
+			getSessionHelper().destroy(sessionName);
+			String scDastUrl = sessionData.getScDastUrlConfig().getUrl();
+			throw new FcliSimpleException(String.format("Unable to connect to SC-DAST URL as configured in SSC; please contact your SSC administrator, or use the --disable option to disable SC-DAST functionality for this session.\nSSC URL: %s\nSC-DAST URL: %s", sscUrl, scDastUrl), e);
         }
     }
 
--- a/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/cmd/SSCSessionLoginCommand.java
+++ b/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/cmd/SSCSessionLoginCommand.java
@@ -27,6 +27,7 @@ import com.fortify.cli.ssc._common.session.helper.ISSCAndScanCentralUrlConfig;
 import com.fortify.cli.ssc._common.session.helper.SSCAndScanCentralSessionDescriptor;
 import com.fortify.cli.ssc._common.session.helper.SSCAndScanCentralSessionHelper;
 
+import kong.unirest.UnirestException;
 import kong.unirest.UnirestInstance;
 import lombok.Getter;
 import picocli.CommandLine.ArgGroup;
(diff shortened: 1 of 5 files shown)


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/cmd/SSCSessionLoginCommand.java
```diff
@@ -59,12 +60,25 @@ protected void testAuthenticatedConnection(String sessionName) {
...
         SSCAndScanCentralSessionDescriptor sessionData = SSCAndScanCentralSessionHelper.instance().get(sessionName, true);
+        String sscUrl = sessionData.getSscUrlConfig().getUrl();
         try ( var unirest = GenericUnirestFactory.createUnirestInstance() ) {
             testAuthenticatedSCSastConnection(unirest, sessionData);
         }
+        catch(UnirestException  expt) {
+        	logoutBeforeNewLogin(sessionName, sessionData);
+        	getSessionHelper().destroy(sessionName);
+        	String scSastUrl = sessionData.getScSastUrlConfig().getUrl();
+        	throw new FcliSimpleException(String.format("SC-SAST is temporarily unavailable. You can still connect to SSC by using the --disable=sc-sast argument.\nSSC URL: %s\nSC-SAST URL: %s" , sscUrl,scSastUrl), expt.getMessage());
```
rsenden: By fcli coding conventions, exception variables should be named `e`, not `expt`; please change.

As for the `FcliSimpleException` that you're throwing:
- You're using the `FcliSimpleException(String fmt, Object... args)` constructor, but given that the output of your `String.format` call doesn't contain any `args` placeholders, `expt.getMessage()` will not be rendered. So, users have no idea why the SC-SAST connection failed (also see next item)
- For consistency with SSC connection failures, we want the full stack trace of the original exception to be shown. So, please pass the full `UnirestException` instance to the `FcliSimpleException` constructor, i.e., use the `FcliSimpleException(String message, Throwable cause)` constructor
- We don't know whether SC-SAST is only 'temporarily' unavailable, and one of the user complaints was that it was unclear as to why fcli was trying to connect to a URL that they didn't specify on the command line. Please change to (something like) one of the following (while still listing actual URLs on next lines):
    - If `--sc-sast-url` was not specified: `Unable to connect to SC-SAST URL as configured in SSC; please contact your SSC administrator, or use the --disable option to disable SC-SAST functionality for this session`
    - If `sc-sast-url` was specified: `Unable to connect to the given SC-SAST URL`

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/cmd/SSCSessionLoginCommand.java
```diff
@@ -59,12 +60,25 @@ protected void testAuthenticatedConnection(String sessionName) {
...
+        }
         try ( var unirest = GenericUnirestFactory.createUnirestInstance() ) {
             testAuthenticatedSCDastConnection(unirest, sessionData);
         }
+        catch(UnirestException  expt) {
```
rsenden: Same comments as above

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/mixin/SSCAndScanCentralSessionLoginOptions.java
```diff
@@ -45,9 +49,26 @@ public static class SSCAndScanCentralUrlConfigOptions extends ConnectionConfigOp
         @MaskValue(sensitivity = LogSensitivityLevel.low, description = "SSC HOST", pattern = MaskValue.URL_HOSTNAME_PATTERN)
         @Getter private String sscUrl;
         
-        @Option(names = {"--sc-sast-url"}, required = false, order=1)
+        @Option(names = {"--sc-sast-url"}, required = false, order=2)
         @MaskValue(sensitivity = LogSensitivityLevel.low, description = "SC-SAST CONTROLLER URL")
         @Getter private String scSastControllerUrl;
+
+        @DisableTest(TestType.MULTI_OPT_PLURAL_NAME)
+		@Option(names = { "--disable" }, required = false, split = ",", order =3, descriptionKey = "fcli.ssc.component.list.disable")
```
rsenden: - Looks like indentation is wrong (at least in GitHub UI). Please make sure that your IDE is configured to use spaces instead of tabs, for example: image
- Please do not use `descriptionKey` on regular options, in order to use default picocli description keys

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/mixin/SSCAndScanCentralSessionLoginOptions.java
```diff
@@ -45,9 +49,26 @@ public static class SSCAndScanCentralUrlConfigOptions extends ConnectionConfigOp
...
+		@Getter private Set<SSCComponentDisable> disable;
```
rsenden: Please rename field and associated getter to something more descriptive like `disabledComponents`. Also, it may be beneficial to initialize this to an empty set, such that callers can assume a non-null value and thus don't need to do null-checks.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/cli/mixin/SSCAndScanCentralSessionLoginOptions.java
```diff
@@ -45,9 +49,26 @@ public static class SSCAndScanCentralUrlConfigOptions extends ConnectionConfigOp
...
+
+		@RequiredArgsConstructor
+		public static enum SSCComponentDisable {
```
rsenden, line 62: Why do you have a `param` field and associated constructor? It's not used anywhere as far as I can see, and even if it's needed for any reason, you could just call `toString()` to generate the same output. Also, in the `toString` method, please use `name()` instead of `super.toString()`. Full enum should look as this:

```java
public static enum SSCComponentDisable {
    sc_sast, sc_dast;

    @Override
    public String toString() {
        return name().replace('_', '-');
    }
}

File: fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/helper/ISSCAndScanCentralUrlConfig.java [hunk 6]
```diff
@@ -3,12 +3,16 @@
...
 import com.fortify.cli.common.rest.unirest.config.IConnectionConfig;
+import com.fortify.cli.ssc._common.session.cli.mixin.SSCAndScanCentralSessionLoginOptions.SSCAndScanCentralUrlConfigOptions.SSCComponentDisable;
 
 /**
  * Interface for the functions to get the SSC URL and the Controller URL
  */
 public interface ISSCAndScanCentralUrlConfig extends IConnectionConfig {
     String getSscUrl();
     String getScSastControllerUrl();
+    Set<SSCComponentDisable> getDisable();
```
rsenden: See earlier comment about renaming the `disable` field to something like `disabledComponents`

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/_common/session/helper/SSCAndScanCentralSessionDescriptor.java
```diff
@@ -112,8 +114,28 @@ public static final SSCAndScanCentralSessionDescriptor create(ISSCAndScanCentral
                 .sscTokenData(sscTokenData)
                 .revokeTokenOnLogout(providedSscToken==null);
         var sscConfigProperties = getSscConfigProperties(sscUrlConfig, sscTokenData.getToken());
-        addScSastSessionConfig(sessionDescriptorBuilder, sscAndScanCentralUrlConfig, sscAndScanCentralCredentialsConfig.getScSastClientAuthToken(), sscConfigProperties);
-        addScDastSessionConfig(sessionDescriptorBuilder, sscAndScanCentralUrlConfig, sscConfigProperties);
+        boolean disableSast = false;
+        boolean disableDast = false;
+        Set<SSCComponentDisable> disabledComponents = sscAndScanCentralUrlConfig.getDisable();
+        if(disabledComponents != null) {
+        	for (SSCComponentDisable disable : disabledComponents) {
```
rsenden: One of the characteristics of a Java `Set` is that it provides an efficient `contains` method. You usually don't iterate over a set to see whether it contains a certain element, but instead just use something like `disabledComponents.contains(SSCComponentDisable.sc_sast)`. Also, temporarily storing this state in separate boolean variables just complicates the code without providing any benefit.

Please remove the boolean variables and the loop for setting those booleans, and instead just do something like:

```java
		if (disabledComponents!=null && disabledComponents.contains(SSCComponentDisable.sc_sast)) {
			sessionDescriptorBuilder.scSastDisabledReason("Disabled by user");
		} else {
                         ...
                }
```

If you make sure that `disabledComponents` can never be `null` (see earlier comment about initializing to empty set), you can even remove the null-check.

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/i18n/SSCMessages.properties
```diff
@@ -82,6 +82,7 @@ fcli.ssc.session.login.token = SSC token in either encoded (REST) or decoded (ap
   use an AutomationToken, CIToken, or UnifiedLoginToken.
 fcli.ssc.session.login.sc-sast-url = Override ScanCentral SAST Controller URL. If not specified, the controller URL as configured in SSC will be used.
 fcli.ssc.session.login.ssc-session = Name for this SSC session. Default value: ${DEFAULT-VALUE}.
+fcli.ssc.session.login.disable = By default, sessions for SSC, SC-SAST, and SC-DAST are created. To bypass a session, use the --disable option.%nAllowed values: ${COMPLETION-CANDIDATES}.
```
rsenden: Only a single session is being created, which by default allows access to all 3 products (if enabled in SSC). Can we please update this to something like the following?

By default, an SSC session allows access to SC-SAST & SC-DAST (if configured). This option allows for disabling SC-SAST and/or SC-DAST for this session, thereby avoiding unnecessary connection checks and preventing any associated failures. Allowed values: ${COMPLETION-CANDIDATES}.

rsenden (changes requested):

Issue #765, opened 2025-06-27
Incorrect generic_action synopsis in online docs


rsenden:
### Current Behavior

The synopsis shown in the online documentation for actions in the `fcli action` module, located at https://fortify.github.io/fcli/latest/generic-actions.html, incorrectly shows `generic_action` as in this screenshot:

![Image](https://github.com/user-attachments/assets/9555cea5-f8f4-416e-aa5f-394f23e6808b)

### Expected Behavior

The online documentation should show the correct synopsis without `generic_action`:

`fcli action run ci [fcli action run options] [action options, see below]`

### Steps To Reproduce

N/A

### Environment

```markdown
N/A
```

### Anything else?

N/A

PR #767, opened 2025-07-02
fix: Incorrect generic_action synopsis in online docs(#765)

jmadhur87:
image


fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/cli/cmd/AbstractActionAsciidocCommand.java
```diff
@@ -72,7 +72,7 @@ public final Integer call() {
     private final String replaceVariables(String s) {
         return s.replace("${version}", FcliBuildPropertiesHelper.getFcliBuildInfo().replace(':', ' '))
                 .replace("${type}", getType())
-                .replace("${typeLower}", getType().toLowerCase());
+                .replace("${actionCmd}", getActionCmd().toLowerCase());
```
rsenden: The return value of `getActionCmd()` should already be in the correct case, so the call to `toLowerCase()` is unnecessary (and if we had any non-lowercase commands like `fcli SSC action`, it would even be incorrect), so please remove `.toLowerCase()`

PR #768, opened 2025-07-02
FCLI Aviator 25.3.0 Release: User Session Token Validation, Remediation Output, and Command Simplifications

kireetivar:
This PR contains following changes:

* **User Session Token Validation**: Added validation to the `aviator session login` command to ensure authentication.
* **Remediation Output**: The `aviator audit` command now generates a `remediations.xml` file with code fix suggestions based on audit results.
* **Command Simplification**: Deprecated the `--email` option in `aviator token delete` and `aviator token revoke` commands, as it is no longer required by the API.
* **Improved Usability**: Made the `--email` option optional for other token commands (e.g., `aviator token list`).
* **Audit Output Cleanup**: Suppressed issues are now excluded from audit results to produce more relevant outputs.

**Note**: **Please Do not merge this PR yet**


fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/_common/session/user/cli/cmd/AviatorUserSessionLoginCommand.java
```diff
@@ -44,6 +56,31 @@ protected void logoutBeforeNewLogin(String sessionName, AviatorUserSessionDescri
...
+                String errorMsg = validationResponse.getErrorMessage();
+                String fullError = "Aviator user token validation failed: " +
+                        (errorMsg == null || errorMsg.isBlank() ? "Token is invalid." : errorMsg);
+                throw new AviatorSimpleException(fullError);
+            }
+            LOG.info("Aviator user token validated successfully with the Aviator server.");
+        } catch (AviatorSimpleException e) {
+            throw e;
+        } catch (Exception e) {
+            LOG.warn("An error occurred while trying to validate the Aviator user token with the server using the 'default' admin config: {}. " +
```
rsenden: What's the rationale for logging a warning, instead of throwing an exception? Usually (for SSC/ScanCentral/FoD), we throw an exception if connection check fails, which will result in a non-zero exit code to inform the caller process like CI pipeline about this failure.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/_common/util/AviatorJwtUtils.java
```diff
@@ -8,41 +8,109 @@
...
 public final class AviatorJwtUtils {
     private static final Logger LOG = LoggerFactory.getLogger(AviatorJwtUtils.class);
     private static final ObjectMapper objectMapper = new ObjectMapper();
 
-    private AviatorJwtUtils() {} // Private constructor for utility class
+    private AviatorJwtUtils() {}
 
-    public static Date extractExpiryDateFromToken(String token) {
-        if (token == null) {
+    private static JsonNode getJwtPayload(String token) {
```
rsenden, line 21: Similar comment as above; why log a warning instead of throwing an exception in case of issues with the JWT token?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
```diff
@@ -42,98 +45,87 @@ public class AviatorSSCAuditCommand extends AbstractSSCJsonNodeOutputCommand imp
     @Option(names = {"--tag-mapping"}, required = false, description = "Tag Mapping") private String tagMapping;
 
     private static final Logger LOG = LoggerFactory.getLogger(AviatorSSCAuditCommand.class);
-    private String auditAction;
+
+    private static final ThreadLocal<JsonNode> cachedResult = new ThreadLocal<>();
```
rsenden: As mentioned before: As per fcli best practices, command classes shouldn’t use instance fields to hold state; only instance fields annotated with Picocli annotations like `@Mixin` or `@Option` are allowed. Although not explicitly stated, storing state in static fields is even worse (and `ThreadLocal` doesn't help); it's not only considered a violation of fcli best practices, but may cause major bugs, especially when used in combination with fcli actions.

As a simple example, see output of the `bulkaudit.yaml` action developed by @fransvanbuul:
image

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/_common/session/user/cli/cmd/AviatorUserSessionLoginCommand.java
kireetivar: This was intended to handle cases like using an older (e.g. 23.2) Aviator server with a newer FCLI version, where validation would always fail but you’re right and i forgot that scenario will actually never happen and I forgot that. I agree I’ll update it to fail fast so it aligns better with our CI/CD expectations.

rsenden: @kireetivar I understand, and in fact having these as warnings allowed me to verify the issue with the static field mentioned below (testing these 25.3 updates against 25.2 Aviator service). Backward compatibility with older Aviator backend versions is also something to consider once we allow customers to run their own Aviator service I guess, so certainly something to keep in mind.

When testing this PR, I think the warning showed something like 'endpoint doesn't exist', so apparently we can differentiate that from other error conditions. So, *if* we'd want to have backward compatibility, we could potentially check for this specific error to identify whether we're talking to Aviator 25.2 or 25.3, and then take the appropriate action (i.e., just ignore this particular error, or output a warning stating that connection check is not supported by current Aviator backend version).

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
kireetivar: it seems I misunderstood how to handle state in command classes. Could you please advise me on the best way to dynamically set the action result in `AviatorSSCAuditCommand` (e.g., `SKIPPED`, `AUDITED`) without using instance fields? I've tried a few other approaches, but they don't seem to be working.

rsenden: Usually, the `getActionCommandResult` method should return a fixed value, which will be considered as the default value for the `Action` column in the table output. Individual records as returned by the `getJsonNode` method may override this default value for the `Action` column by setting a JSON property, for example:
`result.put(IActionCommandResultSupplier.actionFieldName, "SKIPPED")`

PR #795, opened 2025-09-03
feat(aviator): Add auto-remediation from SSC/FoD and advanced audit filtering

kireetivar:
This merge request introduces a significant set of new features and improvements to the `fcli aviator` module for version 25.4.0. The primary goals of this update are to deepen the integration with SSC and FoD for auto remediation, provide powerful new filtering capabilities for the audit process, and introduce automated remediation feature that will apply remediations that are provided by SAST Aviator to local repo of developer.

This release also includes a major optimization to massively improve FPR processing speed.

#### 1. **Advanced SSC Audit Filtering**

The `aviator ssc audit` command has been enhanced with several new options to provide granular control over which issues are processed. This allows users to focus on the most relevant vulnerabilities, significantly speeding up the audit and review process.

  * `--filter-set`: Apply a specific FilterSet by name or ID from SSC.
  * `--ignore-filters`: Bypass all filter sets, including the default one, to process all issues.
  * `--priority`: A powerful option to filter issues directly by priority level (`Critical`, `High`, `Medium`, `Low`), which overrides any other active filter set.
  * `--folder`: Select issues from specific folders within an active FilterSet (e.g., 'Hot', 'Critical').

#### 2. **New Command: `aviator ssc prepare`**

A new utility command has been added to prepare an SSC instance for seamless Aviator integration. This command synchronizes the Aviator custom tags (`Aviator prediction`,`Aviator status`) with specified issue templates and/or application versions. This is a crucial step to prevent SSC from stripping Aviator's audit data when the audited FPR is uploaded.

  * Supports updating a single template/version or all templates/versions on the SSC instance.

#### 3. **New Command: `aviator ssc apply-remediations`**

This new command introduces a powerful "fix-it" capability. It downloads an audited FPR from a specified SSC artifact ID and automatically applies the remediations proposed by SAST Aviator directly to the user's local source code.

  * The target source code directory can be specified with the `--source-dir` option.

#### 4. **New Command: `aviator fod apply-remediations`**

Extending the automated remediation feature to Fortify on Demand (FoD). This command downloads an audited FPR for a given FoD release and applies the proposed code fixes to the user's local source code, mirroring the functionality of the SSC command.

  * The target source code directory can be specified with the `--source-code-directory` option.

#### 5. **Enhanced Application Quota Information**

The `fcli aviator app` commands have been updated to retrieve and display new attributes related to application quota information. The `list`, `get`, `create`, and `update` commands will now show `quota` and `quota_last_updated` fields, providing better visibility into usage.

**Important:**  Do **not** merge this MR until the official release of Core SAST Aviator v25.4.0.


fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/applyRemediation/ApplyAutoRemediationOnSource.java
```diff
@@ -0,0 +1,41 @@
...
+        Path extractedPath;
+        try {
+            extractedPath = ZipUtils.extractZip(file.getPath());
+            LOG.debug("Extracted FPR to path: {}", extractedPath);
+        } catch (IOException e) {
+            LOG.error("Failed to extract FPR file: {}", file.getPath(), e);
+            throw new AviatorTechnicalException("Unable to extract FPR file due to an I/O error.", e);
+        }
+
+        if (!FPRLoadingUtil.hasRemediations(file)) {
```
rsenden: This doesn't seem to depend on the zip file being extracted above, so wouldn't it make more sense to first check whether the FPR file has remediations and only then extract the FPR file, to avoid extracting the FPR file if it doesn't have remediations?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/util/FPRLoadingUtil.java
```diff
@@ -1,90 +1,128 @@
...
 import java.io.File;
 import java.io.IOException;
 import java.util.Enumeration;
 import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+public final class FPRLoadingUtil {
```
rsenden: Having static methods for validating/checking presence of files/extracting files from an FPR file in both this class and `ZipUtils` may result in the FPR file being multiple times. Some examples:
- `validateFpr` method first creates a new `ZipFile` instance to check for presence of `audit.fvdl`, then calls `hasSource` method which creates another `ZipFile` instance to check for source archive presence
- Callers like `ApplyAutoRemediationOnSource` indirectly open the FPR file multiple times; once through `ZipUtils::extractZip`, once through `FPRLoadingUtil::hasRemediations`

I think it would be much better to have a class that represents an FPR file instance, i.e., taking the FPR file path (and/or even `InputStream` that streams the FPR file from SSC/FoD) as a constructor argument, and providing methods to retrieve data from the underlying FPR file. This allows for keeping state/cached data, like readin zip entries only ones to determine which files are present, storing the extracted temp dir directory and cleaning up that directory on instance closure (but also see below), ... An instance of this class would be created only once when then FPR file is being downloaded from FoD/SSC, and passed around to any functionality that needs access to any FPR data.

I also don't really like extracting the FPR file to a temporary directory as this can lead to all sorts of problems (running out of disk space on large FPR files, file access permission issues, files being left behind due to cleanup errors/interruption, ...). If possible, I'd very much prefer to use the Java [Zip File System Provider](https://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/zipfilesystemprovider.html) instead, allowing direct access to zip file contents as if it were a regular file system. We'd need to check for performance implications, especially for large FPR files, but I wouldn't be surprised if zip file system based approach would perform (much) better as it avoids explicit extraction and cleanup of the extracted files.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/audit/AuditFPR.java
```diff
@@ -32,153 +34,102 @@
...
-        ParsedFprData(Map<String, AuditIssue> auditIssueMap, List<Vulnerability> vulnerabilities, FPRInfo fprInfo) {
-            this.auditIssueMap = auditIssueMap;
-            this.vulnerabilities = vulnerabilities;
-            this.fprInfo = fprInfo;
-        }
-    }
-
-
-    public static FPRAuditResult auditFPR(File fprFile, String token, String url, String appVersion, String SSCApplicationName,String SSCApplicationVersion , IAviatorLogger logger, String tagMappingFilePath)
+    public static FPRAuditResult auditFPR(File fprFile, String token, String url, String appVersion,
```
rsenden: Looks like with every update, this method is receiving additional method parameters. Probably good to introduce some additional classes (or reuse existing classes like `SSCAppVersionDescriptor`) to pass around some of this data. For example to combine app version info in a single object, search criteria in another object, ...

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/Vulnerability.java
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * Represents a vulnerability with all extracted data from FVDL.
+ * Implements Searchable for broad, non-specific text searches.
+ * Uses a Builder pattern for safe and robust object creation.
+ */
+@Getter
+@Setter
+@ToString
+public class Vulnerability implements Searchable {
```
rsenden, line 32: Maybe some code reduction is possibly by utilizing some additional Lombok annotations like `@Builder`, `@NoArgsConstructor`, `@RequiredArgsConstructor`, ...?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/applyRemediation/ApplyAutoRemediationOnSource.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+
+public class ApplyAutoRemediationOnSource {
+    private static final Logger LOG = LoggerFactory.getLogger(ApplyAutoRemediationOnSource.class);
+
+    public static int[] applyRemediations(File file, String sourceCodeDirectory, IAviatorLogger logger)
```
rsenden: What does the `int[]` return value represent? Unless not possible for any reason, it would be much better to return a proper result object with proper fields/methods to describe the result.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/fod/cli/cmd/AviatorFODRemediationCommand.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private JsonNode processFprRemediations(UnirestInstance unirest, FoDReleaseDescriptor rd, AviatorLoggerImpl logger) {
+        File downloadedFpr = null;
+        try {
+            logger.progress("Status: Downloading Audited FPR from FOD");
+            downloadedFpr = downloadFprFromFod(unirest, rd);
+            //if(downloadedFpr==null)
+              //  return AviatorFODApplyRemediationHelper.buildResultNode(rd, 0,0,0, "SKIPPED (no FPR available)");
+            logger.progress("Status: Processing FPR with Aviator for Applying Auto Remediations");
+            int[] remediationStatus =  ApplyAutoRemediationOnSource.applyRemediations(downloadedFpr, sourceCodeDirectory, logger);
```
rsenden: Also see comment on ApplyAutoRemediationOnSource about `int[]`; how is anyone readin this code to understand what `remediationStatus[0]` or `remediationStatus[1]` stands for?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/util/StringUtil.java
```diff
@@ -11,6 +11,48 @@ public static boolean isEmpty(String test) {
...
     }
 
+    /**
+     * Strips HTML-like tags from a string to clean it up for display.
+     *
+     * @param description     The string containing tags.
+     * @param stripMarkupTags If true, strips basic formatting tags like <b> and <code>.
+     * @return The cleaned string.
+     */
+    public static String stripTags(String description, boolean stripMarkupTags) {
```
rsenden, line 21: We have similar functionality (but different implementation) for cleaning up issue descriptions in fcli-generated reports, see for example `cleanIssueDescription` method in `ActionSpelFunctions`. Methods in that class are meant to be invoked from fcli actions through SpEL expressions only; these methods shouldn't be invoked from other fcli Java code. However, since we now need similar functionality in multiple places, maybe we should move this code to a different place, such that it can be utilized by both Aviator and fcli actions?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/filter/engine/FilterParser.java
```diff
@@ -0,0 +1,291 @@
...
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class FilterParser {
```
rsenden, line 23: Maybe I'll find out during further review, but can you explain how all new search & filtering functionality in this PR is (planned to be) used (this class, classes in `filter` and `comparer` packages, `Searchable` interface, ...)? 

Fcli already provides generic query functionality based on SpEL expressions, so I want to make sure that we're not implementing similar filter/search functionality in multiple ways. I'm also wondering whether this functionality has been implemented from scratch for the Aviator functionality in fcli, or whether this has been copied/re-used from some existing Fortify library. 

If this code matches SSC/AWB search/filter(set) functionality, can we come up with a better maintainable approach? I've seen way too many inconsistencies in filtering implementations across Fortify products in the past, with a filter/search query entered in AWB showing very different results compared to the same query entered in SSC. Ideally, we should reuse some shared Fortify library that provides the filtering implementation rather than copying/reimplementing this code across multiple products, or maybe we should just use existing SSC API for filtering issues, instead of using client-side filtering on the FPR file?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/fod/cli/cmd/AviatorFODRemediationCommand.java
```diff
@@ -0,0 +1,119 @@
...
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+import picocli.CommandLine.Option;
+
+import java.io.File;
+import java.nio.file.StandardCopyOption;
+
+@Command(name = "apply-remediations")
+public class AviatorFODRemediationCommand extends AbstractFoDJsonNodeOutputCommand implements IRecordTransformer, IActionCommandResultSupplier {
```
rsenden: We try to have class name resemble command name, so would be better to rename class to `AviatorFoDApplyRemediationsCommand`

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCRemediationCommand.java
```diff
@@ -0,0 +1,88 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+import picocli.CommandLine.Option;
+
+import java.io.File;
+
+@Command(name = "apply-remediations")
+public class AviatorSSCRemediationCommand extends AbstractSSCJsonNodeOutputCommand  implements IRecordTransformer, IActionCommandResultSupplier {
```
rsenden: We try to have class name resemble command name, so would be better to rename class to `AviatorSSCApplyRemediationsCommand`

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/fod/cli/cmd/AviatorFODRemediationCommand.java
```diff
@@ -0,0 +1,119 @@
...
+        }finally {
+            if(downloadedFpr!=null && downloadedFpr.exists() && !downloadedFpr.delete())
+                LOG.warn("WARN: Failed to delete temporary downloaded FPR file: {}", downloadedFpr.getAbsolutePath());
+
+        }
+    }
+
+    @SneakyThrows
+    private File downloadFprFromFod(UnirestInstance unirest, FoDReleaseDescriptor releaseDescriptor){
+        File fprFile = File.createTempFile("aviator_" + releaseDescriptor.getReleaseId() + "_" , ".fpr");
```
rsenden: Is this FPR file automatically being deleted upon exit (also upon failure)? Again, would be good to have some generic functionality around FPR file handling encapsulated in a single class, to ensure consistent behavior (including cleanup) across all (Aviator FoD & SSC) commands that handle FPR files for auditing, remediation, ...

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/fod/cli/cmd/AviatorFODRemediationCommand.java
```diff
@@ -0,0 +1,119 @@
...
+        FoDScanDescriptor scanDescriptor = FoDScanHelper.getLatestScanDescriptor(unirest, releaseDescriptor.getReleaseId(),
+                getScanType(), false);
+        FoDScanHelper.validateScanDate(scanDescriptor, FoDScanHelper.MAX_RETENTION_PERIOD);
+        var file = fprFile.getAbsolutePath();
+        GetRequest request = getDownloadRequest(unirest, releaseDescriptor, scanDescriptor);
```
rsenden, line 81: Isn't there any existing functionality in the FoD module for downloading an FPR file that we can reuse here?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/fod/cli/helper/AviatorFODApplyRemediationHelper.java
```diff
@@ -0,0 +1,30 @@
...
+     * Builds the final JSON result node for the command output.
+     * @param rd The SSCAppVersionDescriptor.
+     * @param totalRemediation Total no. of Remediations
+     * @param appliedRemediation Remediations that has been applied successfully
+     * @param skippedRemediation Remediations that has been skipped
+     * @param action Final action.
+     * @return An ObjectNode representing the result.
+     */
+
+    public static ObjectNode buildResultNode(FoDReleaseDescriptor rd, int totalRemediation, int appliedRemediation, int skippedRemediation, String action) {
```
rsenden, line 21: I guess these three separate `int` parameters come from the `int[]` returned elsewhere, but we should really have a proper class like `ApplyRemediationsResult` with separate `int` properties for total/applied/skipped remediation counts (and possibly other info  in the future), instead of an `int[]` where you need to guess which array index represents total/applied/skipped/...

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
```diff
@@ -41,9 +43,12 @@ public class AviatorSSCAuditCommand extends AbstractSSCJsonNodeOutputCommand imp
     @Mixin private ProgressWriterFactoryMixin progressWriterFactoryMixin;
     @Mixin private SSCAppVersionResolverMixin.RequiredOption appVersionResolver;
     @Mixin private AviatorUserSessionDescriptorSupplier sessionDescriptorSupplier;
-    @Option(names = {"--app"}, required = false) private String appName;
-    @Option(names = {"--tag-mapping"}, required = false, description = "Tag Mapping") private String tagMapping;
-
+    @Option(names = {"--app"}) private String appName;
+    @Option(names = {"--tag-mapping"}) private String tagMapping;
+    @Option(names = {"--filterset", "--fs"}) private String filterSetNameOrId;
+    @Option(names = {"--ignore-filters"}) private boolean ignoreFilters;
```
rsenden: According to resource bundle, `--ignore-filters` ignores all filter sets, correct? Can't we merge this into the `--fs` option, like having `--fs none` to not apply any filter sets? If we do need to have a separate option, I think we need a better option name that explicitly mentioned filterset, like `--ignore-filterset` or `--no-filterset`. Otherwise, users might expect this option to ignore the other filtering options like `--priority` as well.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
```diff
@@ -41,9 +43,12 @@ public class AviatorSSCAuditCommand extends AbstractSSCJsonNodeOutputCommand imp
     @Mixin private ProgressWriterFactoryMixin progressWriterFactoryMixin;
     @Mixin private SSCAppVersionResolverMixin.RequiredOption appVersionResolver;
     @Mixin private AviatorUserSessionDescriptorSupplier sessionDescriptorSupplier;
-    @Option(names = {"--app"}, required = false) private String appName;
-    @Option(names = {"--tag-mapping"}, required = false, description = "Tag Mapping") private String tagMapping;
-
+    @Option(names = {"--app"}) private String appName;
+    @Option(names = {"--tag-mapping"}) private String tagMapping;
+    @Option(names = {"--filterset", "--fs"}) private String filterSetNameOrId;
```
rsenden: For SSC, we have `SSCIssueFilterSetResolverMixin` that declares a similar option and provides functionality for loading the filter set descriptor from SSC. Not sure whether this is also applicable here (are filter sets loaded from SSC or FPR?), but to ensure consistent option naming, we should ideally reuse the option definition from the SSC module. Maybe this requires some refactoring of `SSCIssueFilterSetResolverMixin` though, to separate option definition from descriptor loading logic.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
```diff
@@ -41,9 +43,12 @@ public class AviatorSSCAuditCommand extends AbstractSSCJsonNodeOutputCommand imp
...
+    @Option(names = {"--ignore-filters"}) private boolean ignoreFilters;
+    @Option(names = {"--priority", "-p"}, split = ",") @DisableTest(DisableTest.TestType.MULTI_OPT_PLURAL_NAME) private List<String> priorities;
```
rsenden: Still don't like `--priority` and `--folder` as these seem to be somewhat randomly chosen and may cause users to request many additional filtering options like category, file/package, ... Can't we have a generic option like `--query` (taking an SpEL expression like other fcli commands)  or `--filter`/`--search` (taking an SSC/AWB search expression) instead, providing both more flexibility and less maintenance for potentially managing many user-requested filtering options? Alternatively, as filter sets allow for rich filtering options, maybe we should just require users to implement custom filter sets rather than providing filtering options at CLI level

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCPrepareCommand.java
```diff
@@ -0,0 +1,45 @@
...
+import picocli.CommandLine.ArgGroup;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+
+@Command(name = "prepare")
+public class AviatorSSCPrepareCommand extends AbstractSSCJsonNodeOutputCommand implements IActionCommandResultSupplier {
+    @Getter @Mixin private OutputHelperMixins.TableNoQuery outputHelper;
+
+    @ArgGroup(multiplicity = "1..*", heading = "%nUpdate Options (at least one is required):%n")
+    private AviatorSSCPrepareOptionsMixin updateOptions;
```
rsenden, line 21: I don't understand ArgGroup multiplicity, as implementation only handles multiplicity 1, not multiple. Also options seem somewhat weird; update one or all AV/templates, but nothing in between. Need to think about this a bit more; maybe need to have query options, like 'prepare all application versions named test.*', and/or maybe have separate fcli commands; one for preparing appversions, one for preparing templates...

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCRemediationCommand.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            logger.progress("Status: Downloading Audited FPR from SSC");
+            SSCFileTransferHelper.download(
+                    unirest,
+                    SSCUrls.DOWNLOAD_ARTIFACT(ad.getId(), true),
+                    fprFile,
+                    SSCFileTransferHelper.ISSCAddDownloadTokenFunction.ROUTEPARAM_DOWNLOADTOKEN);
+
+            logger.progress("Status: Processing FPR with Aviator for Applying Auto Remediations");
+            int[] remediationStatus =  ApplyAutoRemediationOnSource.applyRemediations(fprFile, sourceCodeDirectory, logger);
```
rsenden: Also see comment on ApplyAutoRemediationOnSource about int[]; how is anyone readin this code to understand what remediationStatus[0] or remediationStatus[1] stands for?

rsenden (changes requested):
I've done a quick review of some key classes, may do further review later on but want to give you something to start with.

Apart from review comments on individual classes, please also fix package structure for helper classes; `helper` packages should be at the same level as the `cli` packages, not underneath the `cli` package:

- `com.fortify.cli.aviator.fod.cli.helper` should be renamed to     
  `com.fortify.cli.aviator.fod.helper`
- `com.fortify.cli.aviator.ssc.cli.helper` should be renamed to     
  `com.fortify.cli.aviator.ssc.helper`

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/fod/cli/cmd/AviatorFODRemediationCommand.java
mneeta: This FPR is getting deleted using below code snippet. 
private JsonNode processFprRemediations(UnirestInstance unirest, FoDReleaseDescriptor rd, AviatorLoggerImpl logger) {
        File downloadedFpr = null;
        try {
            logger.progress("Status: Downloading Audited FPR from FOD");
            downloadedFpr = downloadFprFromFod(unirest, rd);
            logger.progress("Status: Processing FPR with Aviator for Applying Auto Remediations");
            var remediationMetric =  ApplyAutoRemediationOnSource.applyRemediations(downloadedFpr, sourceCodeDirectory, logger);
            String status = remediationMetric.appliedRemediations()>0 ? "Remediation-Applied" : "No-Remediation-Applied";
            return AviatorFoDApplyRemediationsHelper.buildResultNode(rd ,remediationMetric.totalRemediations(), remediationMetric.appliedRemediations(), remediationMetric.skippedRemediations(), status);
        }finally {
            if(downloadedFpr!=null && downloadedFpr.exists() && !downloadedFpr.delete())
                LOG.warn("WARN: Failed to delete temporary downloaded FPR file: {}", downloadedFpr.getAbsolutePath());

        }
    }

So, After we consume the FPR for applying the remediations, we are deleting it using delete() method of File class.

rsenden: I think there might be situations where the `finally` block isn't being executed, like if the user hits `Ctrl-C`. I think it's safer to (also) use [`File::deleteOnExit`](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--)

rsenden: PS. Please use proper markdown syntax (backticks) for quoting code snippets

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/util/FPRLoadingUtil.java
kireetivar: Thanks for the feedback. I agree. We will refactor the FPR handling based on your suggestions:

- We will introduce a new `AutoCloseable` **FPRHandle** class to manage the FPR as a single resource.  
- It will use Java’s Zip File System Provider, eliminating the need for temporary file extraction and its associated risks.  
- **ZipUtils** and **FPRLoadingUtil** will be removed.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
kireetivar: I've renamed it to `--no-filterset` (`--fs none` would be more difficult to find for the user) for clarity and updated the logic to ensure --priority always has precedence.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCPrepareCommand.java
kireetivar: You're right, the current design of the `@ArgGroup` is intentional it handles one update type at a time (e.g., `--update-template <name>` or `--update-all-avs`).

I agree that the one or all approach is a bit rigid. Instead of splitting the command, perhaps we could enhance it to accept comma-separated values for updating multiple specific entities? For example:

*   `--update-template "Template A,Template B"`
*   `--update-av "My App:1.0,Other App:2.0"`

This would provide more granularity while keeping all preparation logic under a single, cohesive command. What do you think of that as a middle-ground solution?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/filter/engine/FilterParser.java
kireetivar: Thanks for raising these points, here’s the thinking behind it:

1. **Why not SpEL?**
   FPR files use a legacy filter syntax (e.g. `[fortify priority order]:Critical`) that doesn’t work with SpEL. We need a parser that understands that language to apply filters correctly.

2. **Is this new code?**
   No , it’s a straight port of the filtering logic from the internal `fortify-model` library (the same one Audit Workbench uses). Porting it was the only way to stay consistent without pulling in closed-source code.

3. **Why not SSC API?**
   The `audit` command filters client-side on the FPR before sending results to Aviator. That flow doesn’t fit with the SSC API.

So while it looks “new,” it’s really just a way to keep `fcli` aligned with Fortify’s standard filtering.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/util/StringUtil.java
kireetivar: That's a great point was not aware that fcli already has this functionality. I'm all for consolidating this into a single, more robust utility.

My main concern is that the `cleanIssueDescription` method uses a proper HTML parsing library, while `stripTags` is regex-based. Their outputs might not be identical, and since that text is used in the llm audit, any difference could potentially impact accuracy.

Given the need for careful testing, what do you think about deferring this to the next release? That would give us a proper window to validate the outputs and ensure we don't introduce any regressions. Thank you!

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
kireetivar: Yeah i agree. A generic `--filter` option would be better in the long run.
For this release, we added `--priority` and `--folder` as simple shortcuts for the most common use cases. Given the tight timeline, building a full filter parser would take time.
What if we keep these for now and I create a ticket to add a more flexible --filter option in the next release?

rsenden: There's already a full filter parser for handling filter sets, right? Doesn't that use similar syntax as AWB/SSC search boxes? Then we can simply use that, allowing users to use something like '--filter 'priority:Critical folder:Hot'`. If that doesn't take too much effort, I'd very much prefer this approach as it's much more flexible and can be fully controlled through a single option. The help output can then display some common examples of how to use the `--filter` option.

Alternatively, having a generic `--query` option like with other fcli commands also shouldn't be too difficult to implement as long as you have some JSON or object representation of vulnerability data; Spring already provides the parsing and evaluation logic; just need to have an object to evaluate the expression on. However, unlike some of the other fcli commands (where users can use something like `... list --output=json` to understand what properties are available for querying), we'd need to list queryable properties somewhere.

Primary issue with postponing introduction of a generic filter/query option is that we can no longer remove the `--priority` and `--folder` options  in fcli v3.x as that's a breaking change, so we can only do so when moving to fcli v4.x. Best we can do in the v3.x range is to mark these options as deprecated, asking users to use the new filter/query option instead. Not too big of a deal for just 2 options, but I'm afraid that generic filter/query option will keep getting postponed due to time constraints, while at the same time, new filtering options are being introduced based on customer request.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/util/StringUtil.java
rsenden: As this is an implementation change, I'm fine with postponing this to the next release. Please create an issue/ticket (preferably on the fcli issue tracker on GitHub) so we don't forget.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/filter/engine/FilterParser.java
rsenden: As this is quite a lot of code, does it have a big impact on fcli .jar/executable size? Is this an exact copy of the relevant code in the `fortify-model` library, or are there any fcli-specific code changes? Does the `fortify-model` library contain any other code that's not used by fcli?

To avoid any inconsistencies, for example forgetting to update the code in fcli if code in `fortify-model` is updated:
- Ideally, the code base that's shared between fcli and `fortify-model` library should be published as a Maven artifact on either Maven Central or some public Fortify/OpenText Maven repository (not sure whether this exists, otherwise we may want to create this ourselves). This way, the filtering code can also remain closed source if necessary, just making the .jar file publicly available.
- For shorter-term solution, we can consider one of the following:
    - Copy fortify-model .jar file (or derived jar file that contains only code needed by fcli) into the fcli repository and dclare this as a dependency
    - Create a separate fcli module (for example named `fortify-model` or `fortify-filter`) containing exact copy of the (relevant) `fortify-model` code, including original package names

We don't want to include more code than necessary, but we do want to re-use existing code as much as possible, rather than copying it into multiple places.

kireetivar:
Hi Ruud, Just to follow up on the latest commit, I've hidden the `aviator fod` command.
The reason for this is that the command's functionality has a dependency on corresponding changes from the FoD side. Since those changes are planned for a future FoD release, I've hidden our command for now to avoid user confusion.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCPrepareCommand.java
rsenden: Even though you'd need to run multiple commands, I think it might be better to have separate commands for preparing templates vs preparing application versions. For example, updating templates might be done by an SSC admin, whereas updating AVs might be done by AV users.

Also, I think it would be good to allow for querying templates/AVs on arbitrary attributes, for example to allow a user to update all AVs for which `createdBy` matches `someUser`. Question is whether we should allow such queries on the `prepare` command(s), or whether users would run something like `fcli ssc av ls --query=... -o 'expr=fcli aviator ssc prepare {id}'`. For comparison, the `fcli ssc av update` command also only supports a single AV; users can use queries on the `list` command to generate multiple update commands. Of course, this is more work for users, so maybe we should revisit this altogether?

Can we get away by allowing only a single template/AV to be updated for now, such that we can decide on bulk updates later? Alternatively, instead of introducing new fcli commands for this, would it be possible to develop an fcli action that utilizes the existing SSC AV/template update commands to implement this functionality?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
kireetivar: As per our discussion. Let’s keep --priority and --folder for now to meet the current release timeline, but ensure that going forward we introduce a generic --filter/--query option for all new filtering requirements. This will avoid the need to add more one-off options and provide users with a more flexible and consistent way to apply filters.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCPrepareCommand.java
kireetivar: As discussed, since we have a requirement to support updating a single application or issue template, we will keep these options for now. Is that okay with you?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
rsenden: Based on the internal discussion, can we please keep only `--folder` and remove `--priority`?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCPrepareCommand.java
rsenden: Lets keep it as is for now. I'm still thinking that it might be good to mark this command as PREVIEW (by adding `(PREVIEW)` in the usage header like we have for some other fcli commands) and explaining in the usage description that that this command is considered preview functionality because it will likely change in a future fcli version to re-use generic tag update functionality that's planned for the `fcli ssc` module.

kireetivar:
@rsenden Added a preflight check before fpr download 
output:
image

PR #796, opened 2025-09-03
Attribute options for microservices

kadraman:
Added options for attributes to be created/updated on microservices and support for `--auto-requried-attrs` option.

Tested on individual microservices and when created as part `fcli fod release create` and `fcli fod action run setup-release`.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/app/attr/helper/FoDAttributeHelper.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/app/attr/helper/FoDAttributeHelper.java
@@ -105,6 +105,15 @@ public class FoDAttributeHelper {
             FoDAttributeDescriptor attributeDescriptor = FoDAttributeHelper.getAttributeDescriptor(unirest, attr.getKey(), true);
             updatesWithId.put(Integer.valueOf(attributeDescriptor.getId()), attr.getValue());
         }
+        if (current == null || current.isEmpty()) {
+            for (Map.Entry<Integer, String> entry : updatesWithId.entrySet()) {
+                ObjectNode attrObj = objectMapper.createObjectNode();
+                attrObj.put("id", entry.getKey());
+                attrObj.put("value", entry.getValue());
+                attrArray.add(attrObj);
+            }
+            return attrArray;
+        };
         for (FoDAttributeDescriptor attr : current) {
             ObjectNode attrObj = objectMapper.createObjectNode();
             attrObj.put("id", attr.getId());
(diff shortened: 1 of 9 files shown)


fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/app/attr/helper/FoDAttributeHelper.java
```diff
@@ -105,6 +105,15 @@ public static JsonNode mergeAttributesNode(UnirestInstance unirest, FoDEnums.Att
             FoDAttributeDescriptor attributeDescriptor = FoDAttributeHelper.getAttributeDescriptor(unirest, attr.getKey(), true);
             updatesWithId.put(Integer.valueOf(attributeDescriptor.getId()), attr.getValue());
         }
+        if (current == null || current.isEmpty()) {
```
rsenden: The full `mergeAttributesNode` method needs some refactoring as it's difficult to understand what it's doing exactly (multiple for-loops on different inputs), where data is being returned (due to multiple `return` statements inside and outside of `if` blocks), and what code is being executed (due to lack of proper `else` block).

Any way to refactor this to get rid of some for loops, and/or having the same logic for handling updates independent of whether current attributes exist or not? If not possible, we should split into multiple methods and user proper if/else statements, like:

```
var updatesById = getUpdatesById(...);
if ( hasCurrentAttributes(...) ) {
  return mergeWithCurrentAttributes(...);
} else {
  return createNewAttributes(...);
}
```
Method names above are for illustration only, probably need much better names to better describe what these methods are doing exactly.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/microservice/cli/cmd/FoDMicroserviceCreateCommand.java
```diff
@@ -40,6 +43,10 @@ public class FoDMicroserviceCreateCommand extends AbstractFoDJsonNodeOutputComma
 
     @Option(names={"--skip-if-exists"})
     private boolean skipIfExists = false;
+    @Option(names={"--auto-required-attrs"}, required = false)
```
rsenden, line 46: Also see https://github.com/fortify/fcli/issues/804

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/app/attr/helper/FoDAttributeHelper.java
kadraman: I have refactored to hopefully make this easier to understand.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/microservice/cli/cmd/FoDMicroserviceCreateCommand.java
kadraman: I will do this refactoring against #804

PR #802, opened 2025-09-09
Fix for the issue `fcli ssc action run ci` fails if Aviator enabled

### LINKED ISSUE #789: `fcli ssc action run ci` fails if Aviator enabled
### Current Behavior

The `fcli ssc action run ci` command fails if Aviator is enabled; see screenshot below. If the `ci` action is run through the top-level action module, like `fcli action run ci` (without `ssc`), the action succeeds.

Image

### Expected Behavior

The `ci` action should succeed if Aviator audit has been enabled, independent of whether it's being run through `fcli action run ci` or `fcli ssc action run ci`.

### Steps To Reproduce

_No response_

### Environment

```markdown

```

### Anything else?

Problem is with the `--aviator-session ${global.sessionName}` option that is being passed to the Aviator login and logout commands; `global.sessionName` is generated by the top-level `ci` action, so this variable resolves to an empty value if the SSC `ci` action is run directly rather than being run through the top-level `ci` action.

Best way to fix, while still using the same session name for SSC & Aviator sessions if both generated through actions, is likely the following:
- Have fcli or the action framework automatically generate a `run id`, which is a UUID that's calculated upon fcli/action startup, and doesn't change even in a chain of nested fcli invocations through `FcliCommandExecutor`
- Make the `run id` accessible to actions through either an action variable or SpEL function, for example `${global.actionRunId}` or `${#action.runId()}`
- Change the `global.sessionName` action variable to a local variable `sessionName` in the top-level `ci` action
- Add the same local `sessionName` variable in the SSC `ci` action

SangameshV:
The `--aviator-session ${global.sessionName}` option used in the Aviator login and logout commands is generated by the top-level CI action. As a result, this variable resolves to an empty value if the SSC CI action is executed directly instead of through the top-level CI action. This has now been updated so that the session name is generated in both the top-level CI action and the SSC action.


fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/runner/ActionRunnerContextSpelFunctions.java
```diff
@@ -22,12 +22,19 @@
...
+
 import lombok.RequiredArgsConstructor;
 
 @Reflectable @RequiredArgsConstructor
 @SpelFunctionPrefix("action.")
 public final class ActionRunnerContextSpelFunctions {
     private final ActionRunnerContext ctx;
+    private final static String RUN_ID = UUID.randomUUID().toString();
+    
+    public final String runID() {
```
rsenden, line 37: Please add SpEL function description

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/ci.yaml
```diff
@@ -44,6 +44,7 @@ steps:
             
   - var.set:
       # Configure scan types to run
+      sessionName: ci-${#action.runID()}
```
rsenden, line 47: Comment in line 46 states 'configure scan types to run'; `sessionName` variable doesn't have anything to do with this. Please reorganize to avoid confusion, for example:

```
- var.set:
    # Configure session name
    sessionName: ...
    # Configure scan types to run
    ...
```
```

**Review by rsenden — CHANGES_REQUESTED** [review]

File: fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/runner/ActionRunnerContextSpelFunctions.java [hunk 3]
```diff
@@ -22,12 +22,21 @@
...
+
 import lombok.RequiredArgsConstructor;
 
 @Reflectable @RequiredArgsConstructor
 @SpelFunctionPrefix("action.")
 public final class ActionRunnerContextSpelFunctions {
     private final ActionRunnerContext ctx;
+    private final static String RUN_ID = UUID.randomUUID().toString();
+    
+    @SpelFunction(cat=workflow, desc="Generates a random UUID, guarantees that each generated id is globally unique.",
```
rsenden: This function description is incorrect; the function doesn't generate a new random UUID on each call. Also, the fact that we use a UUID is an implementation detail; the function description should describe the purpose of the function instead.

Please change to something like the below:
- `desc`: This function returns the current fcli run id, which uniquely represents the current fcli invocation. Different invocations of the fcli executable are guaranteed to have a different, unique run id. Within a single fcli executable invocation, the run id remains the same, even across `run.fcli` instructions and any other internal fcli command invocations.
- `returns`: Current fcli run id in UUID format

fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/runner/ActionRunnerContextSpelFunctions.java
SangameshV: Added the description of the SpEL function.

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/ci.yaml
SangameshV: Reorganised the lines with the addition of another comment to have meaningful comments for the instructions.

fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/runner/ActionRunnerContextSpelFunctions.java
SangameshV: Updated the SpEL Function description.

PR #809, opened 2025-09-29
fix: fcli tool definitions update should validate definitions zip file

jmadhur87:


fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/_common/helper/ToolRegistry.java
```diff
@@ -0,0 +1,21 @@
+package com.fortify.cli.tool._common.helper;
+import java.util.HashSet;
+import java.util.Set;
+
+public class ToolRegistry {
+	private static final Set<String> REGISTERED_TOOLS = new HashSet<>();
+	
+	static {
```
rsenden: Please add comments:

```java
// TODO Whenever support for a given tool is added or removed, update this list
// TODO Any way to have tools register themselves, to avoid this list becoming out of sync with tool commands?
```

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/cli/cmd/ToolDefinitionsUpdateCommand.java
```diff
@@ -19,6 +23,8 @@ public class ToolDefinitionsUpdateCommand extends AbstractOutputCommand implemen
     private static final String DEFAULT_URL = "https://github.com/fortify/tool-definitions/releases/download/v1/tool-definitions.yaml.zip";
     @Getter @Option(names={"-s", "--source"}, required = false, descriptionKey="fcli.tool.definitions.update.definitions-source") 
     private String source = DEFAULT_URL;
+    @JsonIgnore
+    private List<ToolDefinitionsOutputDescriptor> updateToolDefinitions;
```
rsenden: What's the purpose of this field? It doesn't seem to be used anywhere, `@JsonIgnore` doesn't make sense in command classes, and command classes shouldn't have instance fields other than Picocli options/mixins/...

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ZipFileExtractor.java
```diff
@@ -0,0 +1,53 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.FileTime;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+public class ZipFileExtractor {
```
rsenden: I can't find any references to this class; where is this being used? If not being used, please remove.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsZipContent.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Set;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipInputStream;
+import java.nio.file.Files;
+import java.nio.file.attribute.FileTime;
+import java.util.Enumeration;
+import java.util.stream.Collectors;
+
+public class ToolDefinitionsZipContent {
```
rsenden: I can't find any references to this class; where is this being used? If not being used, please remove.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
```diff
@@ -34,104 +38,264 @@
...
+					}
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
+	@SneakyThrows
+	private static void mergeDefinitionsZip(Path dest) {
```
rsenden: Method is way too long and difficult to understand exact behavior

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
```diff
@@ -34,104 +38,264 @@
...
+
+	private static Set<String> getRequiredYamlNames() {
+		Set<String> requiredYamlNames = new HashSet<>();
+		for (String tool : ToolRegistry.getRegisteredToolNames()) {
+			requiredYamlNames.add(tool + ".yaml");
+		}
+		return requiredYamlNames;
+	}
+
+	private static final void addYamlOutputDescriptors(List<ToolDefinitionsOutputDescriptor> result) {
```
rsenden, line 275: Again, method is way too long, making ti difficult to understand exact behavior. Why is there a difference in processing depending on whether `isUpdateDefault` returns true or false, and why is `isUpdateDefault` named like that; it just checks whether `toolDefinitionCustomFilePath` contains `https://`; what has that to do with `default`?

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
```diff
@@ -34,104 +38,264 @@
...
-    }
-    
-}
+	private static final String ZIP_FILE_NAME = "tool-definitions.yaml.zip";
+	public static final Path DEFINITIONS_STATE_DIR = FcliDataHelper.getFcliStatePath().resolve("tool");
+	public static final Path DEFINITIONS_STATE_ZIP = DEFINITIONS_STATE_DIR.resolve(ZIP_FILE_NAME);
+	private static final String DEFINITIONS_INTERNAL_ZIP = "com/fortify/cli/tool/config/" + ZIP_FILE_NAME;
+	private static final Path DESCRIPTOR_PATH = ToolDefinitionsHelper.DEFINITIONS_STATE_DIR.resolve("state.json");
+	private static final ObjectMapper yamlObjectMapper = new ObjectMapper(new YAMLFactory());
+	private static String toolDefinitionCustomFilePath;
```
rsenden: Using a static mutable field like toolDefinitionCustomFilePath is generally considered bad practice. It introduces shared state across all usages of the class, which can lead to race conditions, unexpected side effects, and makes the code harder to test and maintain. Prefer passing such values as method parameters or encapsulating them in an instance field within a properly scoped object.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
```diff
@@ -34,104 +38,264 @@
...
+					Files.copy(filePath, zos);
+					zos.closeEntry();
+					if (latestTime == null || fileTime.compareTo(latestTime) > 0) {
+						latestTime = fileTime;
+					}
+				}
+			}
+		}
+		if (latestTime != null) {
+			Files.setLastModifiedTime(dest, latestTime);
```
rsenden: Looks like you're setting the last modified time of the zip file to the last modified time of the last modified yaml file; why? Shouldn't this be just the current date/time, also with regards to the 'only update if not updated in the past ...' check that we discussed (but which doesn't seem to be part of this PR).

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
```diff
@@ -34,104 +38,264 @@
...
+		}
+
+		for (String yamlFile : yamlFileNames) {
+			Path filePath = DEFINITIONS_STATE_DIR.resolve(yamlFile);
+			try {
+				Files.deleteIfExists(filePath);
```
rsenden: Apparently you're creating temporary files for each YAML file (and possibly also the downloaded zip file, I haven't checked); can you come up with a solution that doesn't require temporary files from being created and cleaned up again (as that's error-prone, for example files being left behind if update is interrupted by user or exception). Of course, we don't want to read the zip-file multiple times, so files may need to be fully processed in the order in which they appear in the zip-file, or you can use a zipfilesystem instead to access files in a zip file as if they were stored on the local file system.

rsenden (changes requested):
Please update your IDE settings to use 4 spaces instead of tabs for indentation, then fix all files in your PR to use spaces instead of tabs.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ZipFileExtractor.java
jmadhur87: Removed

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
jmadhur87: replaced the latest time with current time

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/_common/helper/ToolRegistry.java
jmadhur87: Added these comments, also changed the static block to static method.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/cli/cmd/ToolDefinitionsUpdateCommand.java
jmadhur87: Removed

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsZipContent.java
jmadhur87: Removed this class.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
jmadhur87: Refactored the method.

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
jmadhur87: Instead of isUpdateDefault, directly checking the condition now.
Its verifying if this update command does not contain any source

fcli-core/fcli-tool/src/main/java/com/fortify/cli/tool/definitions/helper/ToolDefinitionsHelper.java
jmadhur87: Resolved

rsenden (commented):
@jmadhur87, can you please review the GitHub Copilot feedback to see whether any of this would be useful to implement/fix?

-----
Useful changes overall — the new behavior (user ZIP support, update modes, "actionResult" reporting) looks reasonable. A few potential bugs, robustness issues and style/maintenance improvements to consider:

Bugs / correctness issues

extractYamlFromZip compares ZipEntry.getName() directly to yamlFile

Problem: entries often have a path (dir/name.yaml). You should compare the file name only (Path.of(entry.getName()).getFileName().toString()) like you do elsewhere. Otherwise the YAML might not be extracted even when present.
Fix: change the equality check to compare filename-only.
NPE when using ZipEntry.getLastModifiedTime()

Several places assume entry.getLastModifiedTime() != null (e.g. addYamlDescriptor, addYamlOutputDescriptors when creating new Date(entry.getLastModifiedTime().toMillis())) — that will NPE if the entry doesn't have a timestamp.
Fix: null-check the LastModifiedTime and use null or a fallback time when it is missing.
getModifiedTime may return null, then new Date(getModifiedTime(dest).toMillis()) will NPE

getModifiedTime now returns null if file doesn't exist. update(...) calls getModifiedTime(dest).toMillis() without checking null. In practice dest should exist, but defensively check for null before using toMillis().
Fix: either ensure getModifiedTime never returns null (throw if missing) or handle the null case in update().
isValidZip only checks whether any required YAML file exists in the zip

You accept zips that contain at least one required yaml. That may allow partial/missing tool definitions. You do later fill missing files from other sources, but make sure the intended behavior is "accept if zip has at least one required yaml" (likely intended) and document it.
Suggestion: if the intention is to accept only fully-contained zips, require all required YAML files.
Path traversal / name sanitization

You usually use Path.of(entry.getName()).getFileName() which mitigates path traversal, but ensure every place writing files/entries uses only the filename (not the full entry name). In mergeDefinitionsZip you write each found file into the output zip with ZipEntry(file.getName()) — good.
MalformedURLException catch logic

You attempt UnirestHelper.download(... new URL(source).toString()) and on MalformedURLException treat source as a local file. Consider other URL issues (e.g., unreachable, 404) that will not throw MalformedURLException but will fail in UnirestHelper.download; ensure failures are handled and have clear error messages.
Exception handling: losing original cause

In several places you throw new FcliSimpleException("Error Loading User ZIP file: " + zipPath) or similar but don't include the original exception as the cause. Include the caught exception so callers can see the stacktrace:
throw new FcliSimpleException("Error Loading User ZIP file: " + zipPath, e);
Regex parsing for duration allows multiple tokens but rejects seconds, which is fine; however the error messages are a bit generic. Consider validating earlier (on CLI) to provide better feedback.

Robustness / resource handling

Good: try-with-resources used in most places. Ensure all streams are closed (e.g., in mergeDefinitionsZip you open InputStream in try-with-resources, fine).
Consider adding more logging (info/debug) for key actions (download, using user ZIP vs state vs internal) to help debugging.
API / design / style suggestions

ToolRegistry list maintenance / dynamic registration
You already leave TODO; consider ServiceLoader or a platform-specific registration mechanism so tools can register themselves and avoid drift between code and registry.
ArgGroup handling in ToolDefinitionsUpdateCommand
You set exclusive = true so picocli should ensure mutual exclusion; explicit runtime check is redundant. It's OK to keep a guard but it's duplicative.
UpdateMode paramLabel strings
paramLabel = "Force Update" is not a typical param placeholder (like <force>); consider removing paramLabel or use a short, meaningful placeholder.
Use clearer error strings and include the failing path or URL and the cause for easier debugging.
Unit tests
Add unit tests for:
parseDurationToDate edge cases and invalid formats
isValidZip with zips that have no matching files / partial matches
mergeDefinitionsZip behavior when some YAMLs are present in user zip and others are taken from internal zip
extractYamlFromZip with entries containing path components
Performance / minor issues

Some methods are long and do multiple responsibilities (mergeDefinitionsZip, addYamlOutputDescriptors). Consider splitting into smaller methods (you already did some splitting — good).
Minor: format and indentation in the diff appears inconsistent; run code formatter / checkstyle.
Security considerations

ZIP handling: always be careful with zip entries that contain large sizes or compressed bombs. You don't seem to unpack arbitrary nested directories (you copy entries directly), but consider bounds checks if working with untrusted zips. Also you currently extract to DEFINITIONS_STATE_DIR; ensure you validate/limit size and number of extracted files.
Small correctness nitpicks

getRequiredYamlFileNames uses ToolRegistry.getRegisteredToolNames().stream().map(s -> s + ".yaml").collect(Collectors.toSet()); fine but consider making ToolRegistry provide the list already suffixed or document expected format.
ToolDefinitionsOutputDescriptor: adding actionResult is good. Ensure consumers of this DTO (JSON serializers / CLI output) handle the added property (you did annotate with JsonProperty using IActionCommandResultSupplier.actionFieldName — good).
Suggested concrete code fixes (high-priority)

In extractYamlFromZip:
change: if (entry.getName().equals(yamlFile)) {
to: if (Path.of(entry.getName()).getFileName().toString().equals(yamlFile)) {
In places that build new Date(entry.getLastModifiedTime().toMillis()), guard:
var lastModified = entry.getLastModifiedTime() != null ? new Date(entry.getLastModifiedTime().toMillis()) : null;
In update(), check getModifiedTime(dest) != null before calling toMillis() (or throw a descriptive exception if it unexpectedly missing).
Include original exceptions when wrapping in FcliSimpleException.
If you want, I can:

Produce a focused patch with the fixes above (extractYamlFromZip filename check, null-checks for lastModifiedTime, include causes in exceptions, fix getModifiedTime usage).
Run a quick search for other places that assume non-null last-modified and produce a complete list to change.
Which of the concrete fixes would you like me to prepare as a patch?

rsenden:
Merged through #870

PR #841, opened 2025-10-27
feat: Re-use common CLI option definitions(804)

jmadhur87:
Moved common options **--auto-required-attrs** and **--skip-if-exists** to CommonOptionMixins.


rsenden:
Build fails due to test failures; please try yourself and fix issues accordingly:

image

...

image

rsenden:
Your latest commit provides a work-around to allow the tests to pass, but doesn't fix the root cause of the failing tests. Also, using shared description keys here doesn't make sense, as different commands that use these options need different descriptions.

rsenden:
@jmadhur87 Please check the visibility of the mixin fields that you added. Looks like you've used `protected` whereas I think most/all of these can be `private`. Also, can you please double-check that there's no risk for NullPointerExceptions? I think picocli always initialized mixin fields (even if none of the options that are defined in the mixin are given by the user), but good to double-check.

PR #844, opened 2025-10-27
fix: Fix SAST reported XXE issues in aviator module

kireetivar:
This PR hardens the application against XXE (XML External Entity) vulnerabilities reported by SAST.

It also includes two related fixes:
* **Bug Fix:** Resolves an `InvalidPathException` in the `apply-remediations` command by correctly normalizing quoted paths and adding path traversal checks.
* **Dependency Update:** Upgrades the `io.grpc:grpc-netty-shaded` library to version `1.76.0`.

### Key Technical Changes
* All instances of `DocumentBuilderFactory`, `TransformerFactory`, and JAXB unmarshallers are now securely configured to disable external entities, DTDs, and secure processing.
* New GraalVM `grpc\reflect-config.json` configs and `grpc\resource-config.json` file are added to support the JAXB and gRPC changes in the native image.


fcli-core/fcli-aviator-common/build.gradle.kts
```diff
@@ -19,7 +19,7 @@ dependencies {
     implementation("com.sun.activation:jakarta.activation:2.0.1")
     implementation("jakarta.xml.ws:jakarta.xml.ws-api:3.0.1")
     implementation("com.auth0:java-jwt:4.5.0")
-    implementation("io.grpc:grpc-netty-shaded:1.69.0")
+    implementation("io.grpc:grpc-netty-shaded:1.76.0")
     implementation("io.grpc:grpc-protobuf:1.69.0")
```
rsenden: Dependency versions for a given library should usually be in sync; can we expect any issues with upgrading grpc-netty-shaded to 1.76.0 while keeping grpc-protobuf as 1.69.0?

kireetivar: agree, upgrading grpc-protobuf to 1.76.0 would also require upgrading the server-side gRPC dependencies to maintain compatibility, which we can only do in the next release.

**Options:**

1. **Keep 1.76.0 for grpc-netty-shaded** - We've tested our scenarios and they work fine, though this technically violates gRPC's version sync best practices
2. **Revert to 1.69.0 for grpc-netty-shaded** - Maintain version consistency now, then address any open-source scan vulnerabilities once the server deployment is done in the next release

Which approach would you prefer?

kireetivar: sorry actually, I was mistaken in my previous response. I've now updated all gRPC dependencies to maintain version consistency.

PR #845, opened 2025-10-28
feat(ssc): Add bulkaudit action for automated SAST Aviator batch auditing of SSC application versions

kireetivar:
This PR introduces a new SSC action, `bulkaudit.yaml`, which automates the process of auditing multiple SSC application versions in bulk using Fortify SAST Aviator. It orchestrates a multi-step workflow, bridging SSC and Aviator to enable efficient, large-scale AI-powered auditing.

### Detailed Workflow

The action executes the following sequence:

1.  **Discover Candidates**: Queries the SSC `/api/v1/issueaging` endpoint to find all application versions with pending audit issues.
2.  **Apply Filters**: Refines the candidate list by applying optional user-provided inclusion (`--filter`) and exclusion (`--exclude-filter`) criteria.
3.  **Provision Aviator Apps**: For each candidate, the action checks if a corresponding application already exists in Aviator. If it does not, the action automatically creates it using `fcli aviator app create`.
4.  **Execute Audit**: Once the Aviator application is confirmed to exist, the action invokes the `fcli aviator ssc audit` command for the specific application version, submitting it for AI-powered analysis.
5.  **Apply Tag Mapping**: A mandatory `--tag-mapping` file is used by the underlying audit command. This is critical for mapping Aviator's `Unsure` results to a specific SSC audit tag, preventing versions from being re-processed indefinitely.

### Key Features & Error Handling

* **Dry-Run Mode**: Includes a `--dry-run` flag to simulate the entire workflow without executing any create or audit commands, allowing for safe validation.
* **Batch Control**: A `--max-audits` flag allows users to limit the number of versions processed in a single run.
* **Entitlement Management**: The action gracefully handles Aviator entitlement exhaustion. If an `aviator app create` command fails (assumed to be due to entitlement limits), it will stop attempting to create *new* applications but will continue to process audits for applications that *already exist* in Aviator.
* **Statistics Tracking**: Provides a summary upon completion, detailing the number of applications created, audits succeeded, audits skipped (due to no auditable issues), and audits failed.

### Prerequisites

Running this action requires three active sessions:

1.  **SSC Session**: To query `issueaging` data and download FPRs.
2.  **Aviator Admin Session**: Required for listing (`app ls`) and creating (`app create`) applications.
3.  **Aviator User Session**: Required for submitting audits via `aviator ssc audit`.

### DIFF
--- a/fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
+++ b/fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
@@ -205,8 +205,6 @@ steps:
           stats.create_successes: 0
           stats.create_failures: 0
           stats.audit_attempts: 0
-          stats.audit_successes: 0
-          stats.audit_failures: 0
           stats.entitlement_exhausted: false
           stats.create_skipped_due_to_entitlement: 0
 
@@ -277,26 +275,12 @@ steps:
                         run_audit:
                           cmd: 'aviator ssc audit --av "${project.project_name}:${project.version_name}" --app "${project.project_name}" --log-level=INFO --tag-mapping="${cli[''tag-mapping'']}"'
                           status.check: false
-                          records.collect: false
 
                     - if: ${cli['tag-mapping'] == null || cli['tag-mapping'] == ''}
                       run.fcli:
                         run_audit:
                           cmd: 'aviator ssc audit --av "${project.project_name}:${project.version_name}" --app "${project.project_name}" --log-level=INFO'
                           status.check: false
-                          records.collect: false
-
-                    # Process results
-                    - if: ${run_audit.exitCode == 0}
-                      steps:
-                        - var.set:
-                            record_count: "${run_audit.records == null ? 0 : run_audit.records.size()}"
-                        - var.set:
-                            record_action: "${(run_audit.records != null && run_audit.records.size() > 0 && run_audit.records[0].action != null && run_audit.records[0].action != '') ? run_audit.records[0].action : 'SKIPPED'}"
-                        - if: ${!(run_audit.records == null || run_audit.records.size() == 0 || run_audit.records[0].action == null || run_audit.records[0].action == '' || (run_audit.records[0].action != null && run_audit.records[0].action.contains('SKIPPED')))}
-                          var.set:
-                            stats.audit_successes: ${stats.audit_successes + 1}
-
 
                     - if: ${run_audit.exitCode != 0}
                       steps:
--- a/fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
+++ b/fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
@@ -1,0 +1,293 @@
+# yaml-language-server: $schema=https://fortify.github.io/fcli/schemas/action/fcli-action-schema-2.1.0.json
+
+author: Fortify
+usage:
+  header: (PREVIEW) Perform SAST Aviator audits of SSC application versions in bulk.
+  description: |
+    This action identifies SSC application versions with pending audit issues and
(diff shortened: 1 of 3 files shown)


fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
```diff
@@ -0,0 +1,328 @@
...
+                # Run audit if app is ready
+                - if: ${app_ready}
+                  steps:
+                    - var.set:
+                        stats.audit_attempts: ${stats.audit_attempts + 1}
+
+                    - if: ${cli['tag-mapping'] != null && cli['tag-mapping'] != ''}
+                      run.fcli:
+                        run_audit:
+                          cmd: 'aviator ssc audit --av "${project.project_name}:${project.version_name}" --app "${project.project_name}" --log-level=INFO --tag-mapping="${cli[''tag-mapping'']}" --refresh=${cli.refresh} --refresh-timeout="${cli[''refresh-timeout'']}"'
```
rsenden, line 286: Please avoid duplicating the command with only difference being the `--tag-mapping` option. For example, you can easily set an `auditCmd` variable in the `var.set` above, then the `run.fcli` steps would use `cmd: '${auditCmd}'` or `cmd: '${auditCmd} --tag-mapping=...'`

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
```diff
@@ -0,0 +1,328 @@
...
+  description: |
+    This action identifies SSC application versions with pending audit issues and
+    automatically submits them for auditing by Fortify SAST Aviator. The action
+    intelligently filters to only process versions with actual pending audit issues,
+    while ignoring versions for which audit information needs to be refreshed.
+
+    For application versions that don't already exist in Aviator, the action will
+    automatically create them before submitting for audit.
+
+    Either --tag-mapping or --add-aviator-tags must be specified. The default tag
```
rsenden, line 15: Custom tag mapping might still leave unsure issues unaudited, and also doesn't guarantee that custom tags referenced in the mapping are properly stored in SSC if those custom tags are not assigned to the appversion.

I think the `audit` command should always check whether all custom tags defined in either default or custom tag mappings have been assigned to the application version, and throw an error if not, as that will cause Aviator audit info to get lost.

Not sure how to best check that current tag mapping doesn't leave unsure issues unaudited, as in general, I guess this is a valid use case for the `audit` command? If that's indeed still a valid use case, maybe we should have an extra (hidden) option on the `audit` command like `--require-audited-status`; if this option is specified (through this bulk-audit action), the `audit` command would check whether all entries in the custom tag mappings will result in the issue to become 'audited', otherwise throw an error.

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/bulkaudit.yaml
```diff
@@ -0,0 +1,328 @@
...
+    names: --filter, -f
+    description: "Optional filter to apply when querying SSC for projects. Example: 'Languages:java'. Default: no filtering"
+    required: false
+    default: ""
+  exclude-filter:
+    names: --exclude-filter, -e
+    description: "Optional inverse filter to exclude projects from audit. Projects matching this filter will be skipped. Example: 'Languages:c#' to exclude .NET projects. Can be combined with --filter. Default: no exclusion"
+    required: false
+    default: ""
+  dry-run:
```
rsenden, line 46: Dry run adds quite a lot of complexity in action implementation, and needs to be explicitly tested. I understand the reason, but is there anything that we can do to make this easier? 

For example, would be be beneficial to have generic support in fcli actions for dry runs? Any suggestions as to what that should look like? For example, maybe we could have a new `run.fcli::skip.if-dry-run` instruction that checks whether the `dry-run` CLI option has been specified, and if so, just outputs something like `Would run: <fcli cmd>`? Any other ideas?

Issue #837, opened 2025-10-22
FCLI SSC CI action does not create application version if SSC not have default issue template


shajaan:
### Enhancement Request

When SSC does not have default issue template, when SSC ci action is run, application version is not created.

When new application version is given in SSC CI action, then,

1.  If default template is set in SSC, use this template to create appversion
2. If only one issue template "In Use" in SSC, use this template to create appversion
3. If many Issue Templates "in Use" and none is set as default, throw an error asking user to set --issue-template with the template name that's available in SSC
Note : If possible , show users the list of available issue templates from the SSC to choose from to enable accessibility

shajaan:
Fixed this issue as part of PR #850

PR #850, opened 2025-10-30
fix: SSC Create application - added In-Use template when no default template (#837)

shajaan:


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue_template/helper/SSCIssueTemplateHelper.java
rsenden, line 1: I think the current implementation can be improved; no need to store the raw JsonNode body, as we already have all descriptors stored in maps. Also, doesn't make much sense to iterate over all templates to find default/single-in-use template if user explicitly provided an `issueTemplateNameOrId`.

Instead, please consider the following implementation:
- Add a new `private final Set<SSCIssueTemplateDescriptors> inUseDescriptors = new HashSet<>();` field
- Update `processIssueTemplate` to add current descriptor to `inUseDescriptors` if `descriptor.isInUse()`
- Change the logic in `getIssueTemplateDescriptorOrDefaultorInUse` method to use proper if-statements (we now need to cover three cases, for which inline operator isn't really suitable); in this order:
    1.  if `isNotBlank(issueTemplateNameOrId)`: `return getDescriptorByNameOrId(...)`
    2. if `defaultIssueTemplateDescriptor!=null`: `return defaultIssueTemplateDescriptor`
    3. if `inUseDescriptors.size()==1`: `return inUseDescriptors.iterator().next()`

PR #862, opened 2025-11-06
feat: Standardize FCLI action option usage in command invocations

jmadhur87:
Changed all SSC and FoD action commands to pass options in quoted format for consistency with run.fcli usage.


fcli-core/fcli-action/src/main/resources/com/fortify/cli/generic_action/actions/zip/ci.yaml
```diff
@@ -95,6 +95,6 @@ steps:
           logout: ${#fcliCmd(logoutEnvPrefix, logoutCmd)} --${module}-session ${sessionName} ${#optsFromEnv(logoutOpts)}
```
rsenden: Looks like you missed this one? Might also want to check for example login command; that's not visible in the PR review page

PR #889, opened 2025-12-16
Redact sensitive information in console logs (stdout & stderr)

### LINKED ISSUE #875: SSC debricked-scan action outputs sensitive data
### Current Behavior

Running the SSC debricked-scan action (either directly or through ci action) will print the Debricked access token in progress messages, for example:

```
START: RUN_DEBRICKED_CLI
Executing fcli tool dcli run -- scan /home/rsenden/workspace/sample-eightball -t TOKEN -r fortify/sample-eightball -b fortify-workflow
SUCCESS: RUN_DEBRICKED_CLI
```

and:

```
START: IMPORT_DEBRICKED_SCAN_REPORT
Executing fcli ssc artifact import-debricked --av fortify/sample-eightball:fortify-workflow -t TOKEN -r fortify/sample-eightball -b fortify-workflow --store debricked_scan_7f83013b_1249_4332_a775_1df806fe59b3
```

### Expected Behavior

All fcli output (like progress messages in this case) should apply masking in a similar way as we currently do for log masking.

We already wrap `System.out` and `System.err` to avoid those from being closed; maybe we can have a wrapper to also automatically replace any sensitive values with appropriate redacted placeholder? 

We might need to capture output line by line though, instead of directly streaming individual characters, which might be an issue with progress messages or other functionality that may not print newline character.

Alternatively, if for example `abcd` needs to be redacted, we could scan all output for `a` first; keep that in a buffer until we receive the next character, then if the next character is `c`, we can safely output `ac`, whereas if next character is `b`, we add it to the buffer and wait for next characters. Once the buffer contains `abcd`, we know it needs to be redacted, so we output the proper redacted placeholder.

Alternatively, we need to look through fcli code to see where potentially sensitive data may be included in the output, and handle redacting those values specifically in those places.

### Steps To Reproduce

Run `fcli ssc action run debricked-scan` or `fcli ssc action run ci` actions.

### Environment

```markdown

```

### Anything else?

_No response_

SangameshV:
fix: Improved console logging by redacting sensitive information from stdout and stderr (fixes #875 )

Please refer to the image below for a sample console output of the `debricked-scan` SSC FCLI action.
image

### DIFF
--- a/fcli-core/fcli-app/src/main/java/com/fortify/cli/app/FortifyCLI.java
+++ b/fcli-core/fcli-app/src/main/java/com/fortify/cli/app/FortifyCLI.java
@@ -13,6 +13,7 @@
 package com.fortify.cli.app;
 
 import com.fortify.cli.app.runner.DefaultFortifyCLIRunner;
+import com.fortify.cli.common.output.transform.mask.StdIoMaskHelper;
 import com.fortify.cli.common.util.ConsoleHelper;
 
 /**
@@ -34,8 +35,10 @@ public class FortifyCLI {
     private static final int execute(String[] args) {
         try {
             ConsoleHelper.installAnsiConsole();
+            StdIoMaskHelper.INSTANCE.install();
             return DefaultFortifyCLIRunner.run(args);
         } finally {
+            StdIoMaskHelper.INSTANCE.uninstall();
             ConsoleHelper.uninstallAnsiConsole();
         }
     }
--- a/fcli-core/fcli-app/src/main/java/com/fortify/cli/app/runner/util/FortifyCLIDynamicInitializer.java
+++ b/fcli-core/fcli-app/src/main/java/com/fortify/cli/app/runner/util/FortifyCLIDynamicInitializer.java
@@ -99,17 +99,8 @@ public final class FortifyCLIDynamicInitializer {
 
     private void registerDefaultLogMaskPatterns() {
         LogMaskHelper.INSTANCE
-            // Log file patterns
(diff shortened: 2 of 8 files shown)


rsenden:
@copilot Please review the masking functionality in this PR, focusing on the following:

- StdioMaskHelper seems to replicate logic that's already available in LogMaskHelper and FortifyCLIDynamicInitializer::registerDefaultLogMaskPatterns; any way to avoid this, potentially by modifying LogMaskHelper to be better suitable for stdio case? Stdio doesn't have log message type, so maybe we need to hold an extra map that contains all mask replacements? Also, are all replacements relevant for stdio, or should we just focus on replacements registered through CLI options and other user-provided data?
- Related to the above, StdioMaskHelper doesn't provide conditional masking based on configured mask level
- Any improvements possible for MaskingPrintStream? For example, why do we pass a new ByteArayOutputStream to the super constructor and then separately save a delegate; can't we just pass delegate to the super constructor and use it from there?
- Any potential interference with generic stdout/stderr processing, AnsiConsole, and existing fcli stdout/stderr wrapping, for example in FcliCommandExecutor?
- Any chance of stdout/stderr becoming corrupted, for example if fcli unexpectedly terminates through kill signal?
- Any way to (further) optimize performance for StdioMaskHelper/LogMaskHelper?
- Any other improvement suggestions?

rsenden:
@copilot Do we really need the stdio-specific patterns that are registered from FortifyCLIDynamicInitializer.java? These seem to be mostly related to potential HTTP responses, not user input. Also, I'm not sure whether any cli options (through MaskValue annotation or through mask setting in fcli action yaml files) are now being properly registered for redaction. For example, debricked-scan.yaml defines debrickedAccessToken with mask settings; these are being registered for log masking but possibly not for stdio masking? We'd want LogMaskingHelper to automatically also add these for stdio masking when being registered for log masking. If there's insufficient information as to whether a value/pattern registered for log masking should also be masked for stdio, assume it needs to be masked in both. If in general there's not sufficient info, just simplify the code to mask all registered values/patterns in both logs and stdio.

PR #936, opened 2026-03-03
Add exponential backoff retry for SSC 502/503 errors addressing temporary server errors without stopping fcli

gilseara:
Add SSCRetryStrategy implementing Apache HttpClient's ServiceUnavailableRetryStrategy to retry on HTTP 502 (Bad Gateway) and 503 (Service Unavailable) with exponential backoff (1s, 2s, 4s) and random jitter (0-500ms). Wire it into the SSC unirest instance via a custom ApacheClient, following the existing FoD retry pattern.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/rest/helper/FoDRetryStrategy.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/rest/helper/FoDRetryStrategy.java
@@ -12,23 +12,30 @@
  */
 package com.fortify.cli.fod._common.rest.helper;
 
+import java.util.concurrent.ThreadLocalRandom;
+
+import org.apache.http.HttpRequest;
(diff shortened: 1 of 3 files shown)


rsenden (commented):
Two questions:
- Is there any risk of POST/PUT/DELETE operations being repeated, causing for example two new entities to be created in SSC rather than just one? Possibly safer to only retry GET requests?
- Should we apply the same retry strategy to SC-SAST/SC-DAST, and maybe also replicate to FoD?

gilseara:
@rsenden 
Agreed, limited 502/503 retry to GET requests; apply to SC-SAST, SC-DAST, and FoD
Only retry GET requests on 502/503 to avoid duplicate side effects from
retrying non-idempotent operations. 
Applied the same retry strategy to SC-SAST and SC-DAST , and added 502/503 GET-only retry
to FoDRetryStrategy alongside the existing 404/429 handling.

PR #943, opened 2026-03-06
fix: enforce FoD app/release exclusivity and correct session login options

SangameshV:
Currently, the `--app` and `--release` options are not defined as mutually exclusive in PicoCLI. Instead, they are manually validated, and an error is thrown when both or neither of these options are provided for the `fcli fod issue list` and `fcli fod oss-scan list-components` commands.

This has now been improved by implementing the PicoCLI approach to define these options as mutually exclusive. As a result, `fcli util all-commands list -o json` will also correctly reflect these exclusive options.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/cli/mixin/FoDAppOrReleaseArgGroup.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/cli/mixin/FoDAppOrReleaseArgGroup.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2021-2026 Open Text.
+ *
+ * The only warranties for products and services of Open Text
+ * and its affiliates and licensors ("Open Text") are as may
+ * be set forth in the express warranty statements accompanying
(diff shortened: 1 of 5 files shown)


rsenden (changes requested):
We're now duplicating `TargetSpecifierArgGroup`, `AppTarget`, `ReleaseTarget` across three different classes; please move this to an ArgGroup implementation that can be re-used across all three classes, with proper naming like `FoDAppOrReleaseArgGroup`.

Also, I noticed that you're extending from `AbstractFoD*Mixin`, but please note that ArgGroups and Mixins are different things, and in general shouldn't be mixed. For example, there might be generic (injection) functionality that operates on Mixins but not ArgGroups, which I guess you already noticed, given the `releaseGroup.setDelimiterMixin()` call.

We don't want each command class having to explicitly set the delimiter mixin, so please research alternative options, for example extending the injection functionality to ArgGroups, or by having an `FoDAppOrReleaseMixin` (which itself is not an ArgGroup, but it may contain the exclusive ArgGroup), with this mixin handling the delimiter mixin injection and selection between app vs release.

PR #944, opened 2026-03-06
fix: correct FoD session login credential grouping and tenant option usage

SangameshV:
The `--tenant` option was not displayed as mandatory for the `fcli fod session login` command when using the `username/password` option group. It is now required when `username` and `password` are provided, but not when using the `client-id` and `client-token` options.


rsenden (changes requested):
This change is not backward-compatible, as pipelines/automations may already be combining `--tenant` with `--client-id` & `--client-secret`; with this change, such pipelines/automations will start failing as this combination is no longer allowed.

We either need to revert to the old behavior, or somehow update the current implementation to *also* accept `--tenant` as an optional/ignored option in combination with `--client-id` and `--client-secret`, for example utilizing Picocli functionality to pre-process  parameters/options and ignoring/removing the `--tenant` option if `--client-id` or `--client-secret` has also been specified.

PR #948, opened 2026-03-17
feat(aviator): Improve apply-remediations UX with --latest, --all-ope…

mneeta:
This enhancement provides a more flexible and user-friendly experience for applying Aviator auto-remediations for SSC by introducing multiple artifact selection modes:
- --latest: Automatically select the most recent Aviator-processed artifact
- --all-open-issues: Process all artifacts with open issues in bulk
- --since: Filter artifacts by upload date (relative: 7d, 2w, 1M; absolute: 2025-01-01) Key Changes:
- Replaced required --artifact-id with flexible selection modes
- Added SinceOptionHelper for robust date/period parsing
- Enhanced SSCArtifactHelper with getLatestAviatorArtifact() and getAllAviatorArtifacts()
- Improved command validation with mutual exclusivity checks
- Added comprehensive unit tests for all new options
- Updated i18n messages with detailed usage descriptions Technical Details:
- SinceOptionHelper supports relative periods (d, w, M, y) and absolute ISO-8601 dates
- DateTimePeriodHelper integration for consistent period parsing across fcli
- Proper UTC timezone handling for date comparisons
- Backward compatible - existing --artifact-id usage unchanged

Enabled apply-remediations command for FoD as well


fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/_main/cli/cmd/AviatorCommands.java
```diff
@@ -28,18 +29,18 @@
...
+                //   command and looks better in the usage command list, as usually
+                //   'rest' has a different header ('Interact with' compared to most
                 //   other commands ('Manage').
                 AviatorAdminConfigCommands.class,
                 AviatorUserSessionCommands.class,
                 AviatorAppCommands.class,
                 AviatorEntitlementCommands.class,
                 AviatorSSCCommands.class,
-//                AviatorFoDCommands.class,
+                AviatorFoDCommands.class,
```
rsenden: There's some discussion as to whether this should live under `fcli aviator` or `fcli fod`. Given that Aviator-related functionality is handled by FoD in the background, none of the other `fcli aviator` commands is relevant for FoD (like user/admin session management, app & entitlement managent, ...). So, unless we plan on supporting any of this functionality also for FoD, moving this to `fcli fod` might make more sense.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -47,49 +52,147 @@
...
+            if (allOpenIssues) {
+                return processAllAviatorArtifacts(unirest, logger, sinceDate);
+            }
+            SSCArtifactDescriptor ad = resolveArtifactDescriptor(unirest, sinceDate);
             return processFprRemediations(unirest, ad, logger);
         }
     }
-    
+
+    private void validateOptions() {
```
rsenden: Please use Picocli ArgGroups for handling exclusivity and related options, either as inner classes in this class, or as a separate single class (with sub-arggroups as inner classes) in the cli.mixin package.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/_main/cli/cmd/AviatorCommands.java
rsenden: As discussed with Frans, since none of the other `fcli aviator` commands make sense for FoD (Aviator user/admin session, app & entitlement management), we'll want to move this to the FoD module, i.e., have `fcli fod aviator apply-remediations`. This likely requires adding a dependency from `fcli-fod` module on `fcli-aviator-common` module, and possibly some refactoring (moving code from `fcli-aviator` to `fcli-aviator-common` if FoD Aviator commands depend on any code in `fcli-aviator` module.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
rsenden: Please use Picocli ArgGroups to model exclusivity and related options, to have Picocli take care of validation and provide consistent error messaging across fcli. The ArgGroups could be either inner classes in this command class, or potentially a single top-level class (with inner classes for sub-arggroups) in the cli.mixin package.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -47,49 +52,147 @@
...
+                appVersionId, artifactsProcessed, artifactsSkipped,
+                totalRemediations, appliedRemediations, skippedRemediations, action);
+    }
+
+    @SneakyThrows
+    private Path downloadArtifactFpr(UnirestInstance unirest, SSCArtifactDescriptor ad, AviatorLoggerImpl logger) {
         Path fprPath = Files.createTempFile("aviator_" + ad.getId() + "_", ".fpr");
-        try (IProgressWriter progressWriter = progressWriterFactoryMixin.create()){
-            logger.progress("Status: Downloading Audited FPR from SSC");
+        try (IProgressWriter progressWriter = progressWriterFactoryMixin.create()) {
```
rsenden: There's already a progress writer instantiated in `getJsonNode` method; nested progress writers are not supported and could cause issues. Also, to avoid having to pass around `unirest`, `logger`, and `progressWriter` all the time, probably better to have an inner class that takes these through Lombok `RequiredArgsConstructor` and performs all of the processing.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -100,7 +203,7 @@ private JsonNode processFprRemediations(UnirestInstance unirest, SSCArtifactDesc
     }
 
     @Override
-    public boolean isSingular() {return true;}
+    public boolean isSingular() { return !allOpenIssues; }
```
rsenden: Fcli command output structure should be consistent, independent of configuration/options. Based on the output being built by `AviatorSSCApplyRemediationsHelper`, it looks like even with `--all-open-issues`, only a single aggregated record is being returned? In that case, `isSingular` should still return `true`, otherwise `false`.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/helper/SinceOptionHelper.java
```diff
@@ -0,0 +1,90 @@
...
+import com.fortify.cli.common.json.JSONDateTimeConverter;
+import com.fortify.cli.common.util.DateTimePeriodHelper;
+import com.fortify.cli.common.util.DateTimePeriodHelper.Period;
+
+/**
+ * Helper for parsing the --since option value into an OffsetDateTime cutoff.
+ * Supports relative durations (e.g. 7d, 2w, 1m, 90d) and absolute date strings
+ * (e.g. 2025-01-01, 2025-01-01T10:30:00, 2025-01-01T10:30:00Z).
+ */
+public final class SinceOptionHelper {
```
rsenden, line 28: Maybe we should move the explicit date string functionality to DateTimePeriodHelper, to support absolute date strings throughout fcli for consistency? We'd need to review existing fcli options that use DateTimePeriodHelper though to see whether combining the existing option name with absolute date string makes sense/is logical.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
```diff
@@ -32,36 +34,139 @@
...
+     * @param appVersionId Application version ID
+     * @param sinceDate    Optional cutoff; only artifacts with uploadDate >= sinceDate are considered. May be null.
+     * @return SSCArtifactDescriptor of the most recent qualifying Aviator artifact
+     * @throws FcliSimpleException if no matching Aviator artifacts found
+     */
+    public static final SSCArtifactDescriptor getLatestAviatorArtifact(UnirestInstance unirest, String appVersionId, OffsetDateTime sinceDate) {
+        // Note: SSC doesn't support filtering by originalFileName in query string,
+        // so we fetch recent artifacts and filter client-side
+        JsonNode response = unirest.get(SSCUrls.PROJECT_VERSION_ARTIFACTS(appVersionId))
+                .queryString("orderby", "uploadDate DESC")
```
rsenden: Looking at the `GET /projectVersions/{parentId}/artifacts` endpoint, it doesn't seem to support `orderBy`, and even if it did, I think SSC usually uses `+` or `-` prefix instead of `DESC` suffix. Anyway, artifacts are returned in descending `uploadDate` order anyway I think.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
```diff
@@ -32,36 +34,139 @@
...
+                .queryString("limit", "50")
+                .queryString("embed", "scans")
+                .asObject(JsonNode.class)
+                .getBody();
+
+        JsonNode data = response.get("data");
+        if (data == null || !data.isArray() || data.size() == 0) {
+            throw new FcliSimpleException(
```
rsenden: Should this throw an exception, or simply return null?

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
```diff
@@ -32,36 +34,139 @@
...
+                "No artifacts found for application version ID: " + appVersionId
+            );
+        }
+
+        for (JsonNode artifact : data) {
+            String originalFileName = artifact.path("originalFileName").asText("");
+            if (!originalFileName.startsWith("aviator_")) { continue; }
+            if (sinceDate != null && !isUploadDateOnOrAfter(artifact, sinceDate)) { continue; }
```
rsenden: Is this logic correct? If upload date for current non-Aviator artifact is older than `sinceDate`, then all subsequent artifacts will also be older than `sinceDate`, so once we reach this point, we'll never find an Aviator artifact that's newer than `sinceDate`.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
```diff
@@ -32,36 +34,139 @@
...
+            return getDescriptor(artifact);
+        }
+
+        throw new FcliSimpleException(buildNoArtifactsMessage(appVersionId, sinceDate));
```
rsenden: Same as above; null or exception?

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
```diff
@@ -32,36 +34,139 @@
...
+     * @throws FcliSimpleException if no matching Aviator artifacts found
+     */
+    public static final List<SSCArtifactDescriptor> getAllAviatorArtifacts(UnirestInstance unirest, String appVersionId, OffsetDateTime sinceDate) {
+        List<SSCArtifactDescriptor> result = new ArrayList<>();
+        int start = 0;
+        int pageSize = 50;
+
+        while (true) {
+            JsonNode response = unirest.get(SSCUrls.PROJECT_VERSION_ARTIFACTS(appVersionId))
+                    .queryString("orderby", "uploadDate ASC")
```
rsenden: Same comment about `orderBy` as above, but given that you're trying to order by ascending date, this might have more impact. Also, if a customer has years of uploads, searching from the oldest date might not be very efficient.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
```diff
@@ -32,36 +34,139 @@
...
+     * Get the latest Aviator-processed artifact for an application version,
+     * optionally filtered to only consider artifacts uploaded on or after sinceDate.
+     *
+     * @param unirest      UnirestInstance
+     * @param appVersionId Application version ID
+     * @param sinceDate    Optional cutoff; only artifacts with uploadDate >= sinceDate are considered. May be null.
+     * @return SSCArtifactDescriptor of the most recent qualifying Aviator artifact
+     * @throws FcliSimpleException if no matching Aviator artifacts found
+     */
+    public static final SSCArtifactDescriptor getLatestAviatorArtifact(UnirestInstance unirest, String appVersionId, OffsetDateTime sinceDate) {
```
rsenden, line 58: Any way to improve this code/have less code duplication between the various Aviator-related methods?

rsenden (changes requested):

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
mneeta: Done! Created a separate AviatorSSCApplyRemediationsArtifactSelectorMixin class in the cli.mixin package with ArtifactSelectionArgGroup as an inner class annotated with @ArgGroup(exclusive = true, multiplicity = "1"), making --artifact-id, --latest, and --all-open-issues mutually exclusive via Picocli.
One constraint worth noting: Picocli does not support @Mixin inside @ArgGroup classes (InitializationException: @Mixins are not supported on @ArgGroups), and placing the same option name (e.g. --appversion) in multiple exclusive sub-groups raises a DuplicateOptionAnnotationsException. To work around this, --av/--appversion and --delim are declared at the top level of the mixin (outside the ArgGroup) with a validate() method enforcing that --av is required with --latest/--all-open-issues and forbidden with --artifact-id. All other exclusivity constraints are fully handled declaratively by Picocli.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
mneeta: Done! Both issues have been addressed:
Nested progress writer removed — downloadArtifactFpr() no longer calls progressWriterFactoryMixin.create(); it now uses the single progressWriter instance created in getJsonNode().
Inner class added — Introduced a private ArtifactProcessor inner class annotated with Lombok's @RequiredArgsConstructor, which holds unirest, logger, and progressWriter as final fields. All processing methods (processAllAviatorArtifacts, processFprRemediations, downloadArtifactFpr) are moved into this class, and the outer class methods access the inner class fields directly — no more parameter passing.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
mneeta: You're absolutely right — regardless of which option is used, AviatorSSCApplyRemediationsHelper always returns a single aggregated ObjectNode (one row summarising totals), never a record per artifact. So the output structure should be consistent and singular across all modes. Fixed — isSingular() now always returns true.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
mneeta: Throwing is the right choice here. The method returns a non-nullable SSCArtifactDescriptor, and the caller (processFprRemediations) uses it directly — returning null would just push the failure further down as an NPE with no context.
This check also catches a distinct failure case: no artifacts exist at all for the app version (e.g. invalid app version ID or nothing ever uploaded), which is different from the check further down that throws when artifacts exist but none are Aviator-processed. Throwing here gives the user an immediate, actionable error message for each scenario.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
mneeta: Great catch, you're absolutely right! Since artifacts are fetched in DESC order (newest first), once we encounter an artifact with uploadDate < sinceDate, all subsequent artifacts will also be older — so continue was wrong and was causing unnecessary iterations. Fixed by:
Checking the date first (before the filename check) — if it's too old, no point checking the filename
Changing continue → break to stop the loop immediately
Also took the opportunity to switch getAllAviatorArtifacts() from ASC to DESC ordering for the same reason — with ASC, a recent --since filter would wastefully fetch all older pages before reaching matching artifacts. With DESC, we hit the cutoff early, break, then reverse the result list before returning to maintain the ascending order contract.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/artifact/helper/SSCArtifactHelper.java
mneeta: Done! Extracted the shared logic into three private helpers:
fetchAviatorArtifacts(unirest, appVersionId, sinceDate, maxResults) — single unified method handling DESC fetch, Aviator filename filtering, date-based early termination, and pagination; maxResults = 1 gives getLatestAviatorArtifact() behaviour, Integer.MAX_VALUE gives getAllAviatorArtifacts() behaviour
isAviatorArtifact(artifact) — extracts the aviator_ filename check
shouldStopProcessing(artifact, sinceDate) — extracts the date cutoff early-termination check
Both public methods are now thin wrappers — getLatestAviatorArtifact() calls fetchAviatorArtifacts(..., 1) and returns the first result, while getAllAviatorArtifacts() calls fetchAviatorArtifacts(..., MAX_VALUE) and reverses before returning to maintain the ascending order contract.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/_main/cli/cmd/AviatorCommands.java
mneeta: I'd lean towards keeping this under fcli aviator fod and disagree with moving it to fcli fod, for the following reasons:

1. Feature ownership is Aviator, not FoD. The command's primary purpose is applying AI-generated code remediations to source files using Aviator's engine (ApplyAutoRemediationOnSource). FoD's only role is providing the FPR file download — it's the data source, not the product context. The same logic would apply to fcli aviator ssc apply-remediations (it also just downloads an FPR from SSC), and we wouldn't move that to fcli ssc.

2. Symmetry and discoverability. Having fcli aviator ssc apply-remediations and fcli aviator fod apply-remediations as parallel commands gives users a consistent mental model — regardless of which product stores their scans, the Aviator remediation experience lives in the same place. A developer thinking "I want Aviator to fix my code" will look under fcli aviator, not fcli fod.

3. It's true that most fcli aviator commands (user/admin session, app, entitlement management) are SSC-specific and not applicable to FoD. But apply-remediations is precisely the one Aviator feature that IS relevant to FoD users — it's the core end-user-facing capability. That distinction is a reason to grow fcli aviator fod further, not to move this command out.

Happy to discuss further if there's a strong reason for the alternative, but the current placement feels architecturally correct.

PR #960, opened 2026-04-01
feat: Add --force-delete option to sc-dast scan delete

gilseara:
The goal is to match the functionality offered on the GUI for deleting scans that are problematic (i.e. FailedToPause, etc...), see ForceDelete option below: 
 
image

## Summary
- Adds `--force-delete` / `-f` option to `fcli sc-dast scan delete` command
- When specified, appends `forceDelete=true` query parameter to the scan-action API call (`/api/v2/scans/<id>?forceDelete=true`)
- Adds a protected `updateRequest()` hook in `AbstractSCDastScanActionCommand` so subclasses can customize HTTP requests

## Test plan
- [ ] Run `fcli sc-dast scan delete <id>` without `--force-delete` — should behave as before
- [ ] Run `fcli sc-dast scan delete <id> --force-delete` — should include `?forceDelete=true` in the API call
- [ ] Run `fcli sc-dast scan delete <id> -f` — same as above using short flag
- [ ] Verify other scan action commands (pause, resume, complete, etc.) are unaffected


rsenden:
Merged with changes in 0fb8a4dbb249d008551b4c4850c215f5f123b736

Issue #839, opened 2025-10-23
FCLI - SSC create custom tag - does not have option to select if the LIST tag is extensible or not


shajaan:
### Enhancement Request

 Custom tag always creates an extensible list tag
 User should be able to select if the tag should be extensible or not

- Create a fcli ssc session
- Run command fcli ssc tag create <tagname> --type LIST --values=<values>

PR #961, opened 2026-04-06
Added extensible option in ssc tag create and update command

jmadhur87:
feat: Added extensible option in ssc tag create and update command([fix 839](https://github.com/fortify/fcli/issues/839))

User can set extensible flag for creating list type ssc custom tag and also can update.

Extensible allow to add new values to list type custom tag while remediation.


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/custom_tag/cli/cmd/SSCCustomTagUpdateCommand.java
```diff
@@ -52,6 +52,8 @@ public class SSCCustomTagUpdateCommand extends AbstractSSCJsonNodeOutputCommand
     private Boolean hidden;
     @Option(names = {"--requires-comment"}, required = false, negatable = true)
     private Boolean requiresComment;
+    @Option(names = {"--extensible"}, required = false)
```
rsenden: For consistency, I think we should have `negatable = true` here as well; please implement and check for expected behavior

rsenden (changes requested):

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/custom_tag/cli/cmd/SSCCustomTagUpdateCommand.java
jmadhur87: Added negatable = true, for this option

Issue #959, opened 2026-03-31
Support Adding New Values to Extensible List-Type Custom Tags in SSC Issue Update Command


jmadhur87:
### Enhancement Request

When updating SSC issues with a LIST type custom tag marked as extensible, add an optional flag to allow users to automatically create new tag values if they don't exist. The system should validate that the tag is both LIST type and extensible before creating new values;

PR #972, opened 2026-04-13
Support adding new values to extensible list-type custom tags in SSC issue update during remediation

jmadhur87:
This PR enhances the SSC issue update command used during remediation to support adding new values to custom tags of type list when the tag is marked as extensible=true using new option --extend.

While verifying for` ssc tag update` command, found that value updates for non-LIST custom tags (TEXT, DECIMAL, DATE) were incorrectly reported as Updated in fcli, even though SSC made no change.
Added explicit validation and a clear error message so these unsupported value operations now fail fast instead of showing a false success.

Fix: Support adding new values to extensible list-type custom tags in SSC issue update during remediation(#959 )


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/custom_tag/helper/SSCCustomTagUpdateHelper.java
rsenden, line 1: Although the class name is somewhat misleading, originally this class only contained helper methods for obtaining the list of custom tags to be assigned/removed to an app version or issue template; it didn't actually do any updates on SSC. The new functionality added in this PR is completely unrelated to the original functionality.

Please refactor:
- Rename the original `SSCCustomTagUpdateHelper` class (without your changes) to `SSCCustomTagAssignmentHelper` for better clarity
- Rename the original `SSCCustomTagHelper` class to `SSCCustomTagDefinitionHelper` for better clarity
- Move your code for updating custom tag values to `SSCCustomTagDefinitionHelper`
- Look for opportunities for better code reuse; for example, if it makes sense, utilizing `SSCCustomTagDefinitionHelper` for creating/updating a custom tag definition, instead of having this implemented directly in the SSCCustomTagCreate/UpdateCommands

rsenden (changes requested):

Issue #974, opened 2026-04-15
Add `get` command for `fcli ssc issue` and `fcli fod issue`


jmadhur87:
### Enhancement Request

Currently, `fcli ssc issue list` and `fcli fod issue list` support listing issues, but there is no dedicated `get` command to fetch a single issue by ID. A `get` command would allow users and action workflows to efficiently retrieve a specific issue without listing all issues and filtering client-side. Add a `get` command to both `fcli ssc issue` and `fcli fod issue` that retrieves a single issue by its identifier, with optional `--embed` support to include additional data in the response.

### SSC
- `fcli ssc issue get <issue-id> --av <appversion> [--embed details|comments|auditHistory]`
   - `<issue-id>`: The numeric `id` field from the SSC issue response
   - `--embed`: Optional; values from existing `SSCIssueBulkEmbedMixin` (`details`, `comments`, `auditHistory`)
   - REST endpoint: `GET /api/v1/projectVersions/{id}/issues/{issueId}` or resolve via list + filter

### FoD
- `fcli fod issue get <vuln-id> --rel <release> [--embed details|recommendations|traces|...]`
  - `<vuln-id>`: The `vulnId` field from the FoD vulnerability response
  - `--embed`: Optional; values from existing `FoDIssueEmbedMixin` (`details`, `recommendations`, `traces`, `history`, `requestResponse`, `headers`, `parameters`, `allData`, `summary`)
  - REST endpoint: `GET /api/v3/releases/{releaseId}/vulnerabilities/{vulnId}`

## Implementation Notes
- Follow the existing `FoDAppGetCommand` / `SSCAppVersionGetCommand` patterns
- Use `FoDIssueEmbedMixin` (FoD) and `SSCIssueBulkEmbedMixin` (SSC) for `--embed` option
- Extend `AbstractFoDJsonNodeOutputCommand` (FoD) and `AbstractSSCOutputCommand` (SSC)
- Register commands in the respective parent `IssueCommands` container class
- Add usage header/description to `FoDMessages.properties` and `SSCMessages.properties`
- For SSC: requires `--appversion` / `--av` to resolve the `projectVersionId`
- For FoD: requires `--release` / `--rel` to resolve the `releaseId`

## Acceptance Criteria

- [ ] `fcli ssc issue get <id> --av <appversion>` returns the single issue JSON
- [ ] `fcli fod issue get <vuln-id> --rel <release>` returns the single vulnerability JSON
[...]

PR #978, opened 2026-04-17
Add issue get command for FoD and SSC

jmadhur87:
Added issue get command for both SSC and FoD to fetch a single issue by ID, avoiding full list + client-side filtering.
Supports optional --embed to include additional issue details.

fcli fod issue get
fcli ssc issue get

feat: Add get command for fcli ssc issue and fcli fod issue(#974 )


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    protected IObjectNodeProducer getObjectNodeProducer(UnirestInstance unirest) {
+        String appVersionId = parentResolver.getAppVersionId(unirest);
+        return requestObjectNodeProducerBuilder(ObjectNodeProducerApplyFrom.SPEC)
+                .baseRequest(getBaseRequest(unirest, appVersionId))
+                .build();
+    }
+
+    private HttpRequest<?> getBaseRequest(UnirestInstance unirest, String appVersionId) {
+        return unirest.get(SSCUrls.PROJECT_VERSION_ISSUE(appVersionId, id)).queryString("qm", "issues");
```
rsenden: Please double-check, but I think `qm` request parameter is not supported by SSC on this endpoint.

Also, we could consider using the issueDetails endpoint instead as that doesn't require application version to be specified (thereby simplifying CLI usage), but output of that endpoint is not the same as what we return on the `list` command, so users may not expect that inconsistency.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,87 @@
...
+        String releaseId = releaseResolver.getReleaseId(unirest);
+        JsonNode issue = getIssue(unirest, releaseId);
+        return simpleObjectNodeProducerBuilder(ObjectNodeProducerApplyFrom.SPEC)
+                .source(issue)
+                .build();
+    }
+
+    private JsonNode getIssue(UnirestInstance unirest, String releaseId) {
+        boolean numericId = vulnId!=null && vulnId.chars().allMatch(Character::isDigit);
+        JsonNode issue = numericId
```
rsenden: I don't understand this logic; if an issue is numeric, we first try filtering on `id` and if not found, filter on `vulnId`, and vice versa if not numeric? Is there any way to identify which of these two has been specified by the user before making the REST call?

(but also see generic comment about what id's we should actually support on this command)

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,87 @@
...
+            throw new FcliSimpleException(String.format("No issue found for id or vulnId '%s' in the specified release", vulnId));
+        }
+        return issue;
+    }
+
+    private JsonNode getIssueByFilter(UnirestInstance unirest, String releaseId, String fieldName, String value) {
+        HttpRequest<?> request = unirest.get(FoDUrls.VULNERABILITIES)
+                .routeParam("relId", releaseId)
+                .queryString("filters", fieldName+":"+value)
+                .queryString("limit", "1");
```
rsenden: In addition to the Copilot comment, when using a 'list' endpoint with a filter that should result in only a single record being returned, it's better to specify `limit=2` to ensure that the endpoint indeed only returned a single item. See https://github.com/fortify/fcli/blob/f64243ebaeb69f7581114e5912b05a2bd0f42ccc/fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/appversion/helper/SSCAppVersionHelper.java#L67 and corresponding `versions.size()>1` check in `getOptionalDescriptor` method just below.

rsenden (changes requested):
Apart from the individual comments on implementation details, we need to decide on what id's we should support, which is a tradeoff between fcli consistency vs usability.

For (almost) all other `get` commands in fcli, we just accept a single type of id, which corresponds to the id expected by the corresponding REST endpoint. This is what you've implemented for SSC, and would simplify the FoD implementation.

Also, we want to ensure consistency between `issue get` and `issue update` commands; these should accept the same type(s) of id('s). I think both FoD and SSC `issue update` commands currently only accept a single type of id (`id` for SSC, `vulnId` for FoD), so we should have the `issue get` commands accept those same types of id's only.

Having said that, for future consideration, we might want to consider deviating from this approach given that issues have multiple id's (id, issue instance id, and for FoD, vulnId). 

For example, although users can see the SSC issue id in the output of the `issue list` command, I don't think this id is visible anywhere in the SSC UI. So, if users view an issue in SSC and then want to work on this issue in fcli (through `issue get` or `issue update` commands), they need to somehow figure out the issue id. Even SSC itself uses issueInstanceId filter instead of passing the issue id when opening the details for a given issue in a different browser window.

We can leave support for multiple types of id's as a future enhancement, although we need to think about the potential impact on command structure, for example whether we should have explicit options instead of a positional parameter (but that's inconsistent with other fcli `get` commands), to be able to differentiate between `--id` and `--instance-id` for example if we can't automatically determine the type of id being queried by the user. Alternatively, we can just try passing the given id through multiple REST queries on different fields (like you're more or less doing for FoD now, but not optimal from performance perspective), or support an optional prefix for the positional parameter, i.e., `fcli ssc issue get instance-id:...` vs `fcli ssc issue get id:...`.

Somewhat related, for the `issue update` commands, I can imagine having a generic query option, i.e., update all issues that match a given query to perform bulk update, for example updating all issues in SQL Injection category, but again that's a potential future enhancement.

To summarize, maybe best to just have the `issue get` commands accept the same type of id as accepted by the `issue update` command for now, and raising an enhancement request for supporting additional id types on the `issue get` and `issue update` commands (again, these should be consistent as to what id types they support, likely through some shared helper class or resolver mixin), and another enhancement request for adding query support on the `issue update` command (maybe such an ER already exists though).

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,87 @@
...
+import kong.unirest.HttpRequest;
+import kong.unirest.UnirestInstance;
+import lombok.Getter;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+import picocli.CommandLine.Parameters;
+
+@DisableTest(TestType.CMD_DEFAULT_TABLE_OPTIONS_PRESENT)
+@Command(name = OutputHelperMixins.Get.CMD_NAME)
+public class FoDIssueGetCommand extends AbstractFoDOutputCommand {
```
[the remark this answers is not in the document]
jmadhur87: Removed the disable test, added yaml output as default and other output format, table and json also.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: Added changes to supports Vulnerability Id only.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
jmadhur87: Verified the SSC api reference , query parameter is available for this endpoint.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: changed the limit to 2

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
rsenden: I think this uses the following endpoint, which doesn't list a `qm` query parameter:

Image

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,89 @@
...
+                .build();
+    }
+
+    private JsonNode findIssue(UnirestInstance unirest, String releaseId) {
+        HttpRequest<?> request = unirest.get(FoDUrls.VULNERABILITIES)
+                .routeParam("relId", releaseId)
+                .queryString("filters", "vulnId:" + vulnId)
+                .queryString("limit", "2");
+        var response = includeMixin.updateRequest(request).asObject(JsonNode.class);
+        if ( response.getStatus() >= 400 ) {
```
rsenden: Why do you explicitly check response status? Shouldn't fcli already throw an error on FoD error responses?

rsenden (changes requested):
Please check review comments. Also, as there's been some refactoring on FoD issue-related classes in http-mcp branch (which has higher priority to be merged than this PR), let's wait until that branch has been merged to dev/v3.x first, then update this PR as needed before merging it.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: added the changes

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
```diff
@@ -0,0 +1,56 @@
...
+import picocli.CommandLine.Parameters;
+
+@Command(name = OutputHelperMixins.Get.CMD_NAME)
+public class SSCIssueGetCommand extends AbstractSSCOutputCommand {
+    @Getter @Mixin private OutputHelperMixins.Get outputHelper;
+    @Mixin private SSCAppVersionResolverMixin.RequiredOption parentResolver;
+    @Parameters(index = "0", arity = "1", descriptionKey = "fcli.ssc.issue.get.id")
+    private String id;
+    @Mixin private SSCIssueBulkEmbedMixin bulkEmbedMixin;
+    @Mixin private SSCIssueIncludeMixin includeMixin;
```
rsenden: `SSCIssueIncludeMixin` doesn't make sense on a `get` command; if you specify an issue id, we should just get that id independent of whether the issue is visible, removed, ... Also, similar to my earlier comment below, this command uses the following SSC endpoint:

Image

This endpoint supports neither the `showhidden`/`showremoved`/`showsuppressed` request parameters that are generated by `SSCIssueIncludeMixin`, nor the `qm` request parameter for which I raised a comment earlier.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
rsenden: Code still shows `qm` request parameter, which is not supported on `GET
/projectVersions/{parentId}/issues/{id}` endpoint.

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+@Command(name = OutputHelperMixins.Get.CMD_NAME)
+public class FoDIssueGetCommand extends AbstractFoDOutputCommand {
+    @Getter @Mixin private OutputHelperMixins.Get outputHelper;
+    @Mixin private FoDDelimiterMixin delimiterMixin; // Is automatically injected in resolver mixins
+    @Mixin private FoDReleaseByQualifiedNameOrIdResolverMixin.RequiredOption releaseResolver;
+    @Parameters(index = "0", arity = "1", descriptionKey = "fcli.fod.issue.get.vulnId")
+    private String vulnId;
+    @Mixin private FoDIssueEmbedMixin embedMixin;
+    @Mixin private FoDIssueIncludeMixin includeMixin;
```
rsenden: Actually, same is true here as what I wrote for SSC. Although (contrary to SSC) the FoD endpoint used by this command does support the query parameters generated `FoDIssueIncludeMixin`, the user shouldn't need to explicitly pass `--include` if the given issue id is hidden/removed/... If user passes an id, we should just gt that id regardless of issue state. In other words, we should automatically pass query parameters to find issues in any state.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
jmadhur87: For this, using the endpoint as of list for the get command as well. Because we want embed feature also to have all the information about the issue.
image

This api support showhidden/showremoved/showsuppressed

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: To confirm my understanding  the fix here would be to remove the [--include] option from the [get] command and hardcode includeFixed=true and includeSuppressed=true in [findIssue()] for both ssc and fod, so that any issue can be retrieved by ID regardless of its state, without requiring the user to pass extra flags. Is that what you're expecting?

rsenden: Correct for FoD; you'd always add the request parameters to have FoD return fixed/suppressed/... issues, such that FoD will return the issue data for the requested id independent of whether it's visible, fixed, suppressed, ...

For SSC, you're using an endpoint to get the issue data for a specific issue id. This endpoint will return that data independent of whether the issue is hidden/removed/suppressed; the endpoint doesn't support the request parameters generated by the `SSCIssueIncludeMixin`, and doesn't support the `qm` request parameter as explained in other comments.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
rsenden: No, your code is not using the endpoint shown in your screenshot:

- `list` command uses `/projectVersions/{parentId}/issues`
- `get` command uses `/projectVersions/{parentId}/issues/{id}` (through `SSCUrls.PROJECT_VERSION_ISSUE(appVersionId, id)`)

Although very similar, the latter adds the issue id at the end of the path, so these are different endpoints with different behavior and a different set of supported request parameters.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/helper/FoDIssueHelper.java
```diff
@@ -105,25 +105,29 @@ private String asString(Set<String> values) {
     /** Overload adding aggregation fields to an ObjectNode using provided data. */
     public static final ObjectNode transformRecord(ObjectNode record, IssueAggregationData data) {
```
rsenden, line 106: I don't understand how the code changes in this method are related to the goal of this PR (adding `issue get` command), and current `toJsonNode` implementation is wrong. Data types in fcli output must be stable to allow for external processing, i.e., not switch between array or text node depending on the number of available entries. Also, in general, `N/A` should be modelled as `null` or empty array, not `N/A` text node as that's difficult to differentiate from actual values.

So, all the array fields must be either a proper array (which may be empty, single value, or multiple values array), or possibly `null`, like it was in the original code.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
```diff
@@ -0,0 +1,72 @@
...
+    protected IObjectNodeProducer getObjectNodeProducer(UnirestInstance unirest) {
+        String appVersionId = parentResolver.getAppVersionId(unirest);
+        return requestObjectNodeProducerBuilder(ObjectNodeProducerApplyFrom.SPEC)
+                .baseRequest(getBaseRequest(unirest, appVersionId))
+                .build();
+    }
+
+    private HttpRequest<?> getBaseRequest(UnirestInstance unirest, String appVersionId) {
+        return unirest.get(SSCUrls.PROJECT_VERSION_ISSUE(appVersionId, id))
+                .queryString("showHidden", "true")
```
rsenden, line 53: I think the `PROJECT_VERSION_ISSUE` endpoint doesn't support these query parameters; it will always return the issue that corresponds to the given id, independent of whether that issue is hidden, removed, or suppressed.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueGetCommand.java
```diff
@@ -0,0 +1,72 @@
...
+                .queryString("showRemoved", "true")
+                .queryString("showSuppressed", "true");
+    }
+
+    @Override
+    public HttpRequest<?> updateRequest(HttpRequest<?> request) {
```
rsenden, line 59: I don't understand this code:
- The `PROJECT_VERSION_ISSUE` endpoint used by this command doesn't support the `qm` query parameter
- Even if `qm` was supported, I don't see why `embedNames` would need to be appended to its value; embed names are fcli-specific identifiers that SSC doesn't have any knowledge about (apart from that they may overlap with SSC endpoint names, but even then still not valid values for `qm` request parameter).

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueEmbedderSupplier.java
```diff
@@ -39,49 +39,74 @@ public ISSCEntityEmbedder createEntityEmbedder() {
...
-        }
+        protected abstract String getEmbedFieldName();
 
         protected abstract HttpRequest<?> getBaseRequest(UnirestInstance unirest);
+
         protected abstract void process(ObjectNode record, JsonNode response);
+
+        @Override
+        public final void addEmbedRequests(SSCBulkRequestBuilder builder, UnirestInstance unirest, JsonNode record) {
+            // Skip if data is already present in the record (e.g. returned by SSC via qm parameter)
```
rsenden, line 50: This assumption is nonsense; SSC `qm` parameter (which is not even supported on the endpoint that's used by the `issue get` command) doesn't control embedding of extra data. Please undo all irrelevant code changes to this class and all related classes (i.e., remove all `getEmbedFieldName` methods)

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,85 @@
...
+    @Parameters(index = "0", arity = "1", descriptionKey = "fcli.fod.issue.get.vulnId")
+    private String vulnId;
+    @Mixin private FoDIssueEmbedMixin embedMixin;
+
+    @Override
+    protected IObjectNodeProducer getObjectNodeProducer(UnirestInstance unirest) {
+        FoDReleaseDescriptor releaseDescriptor = releaseResolver.getReleaseDescriptor(unirest);
+        String releaseId = releaseDescriptor.getReleaseId().toString();
+        JsonNode issue = findIssue(unirest, releaseId);
+        if ( issue==null ) {
```
rsenden: I think it makes more sense to move this null-check into the `findIssue` method; that method already throws an exception if multiple issues were found, so it's logical to make that method also responsible for throwing an exception if no issues were found.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueEmbedderSupplier.java
```diff
@@ -39,49 +39,53 @@ public ISSCEntityEmbedder createEntityEmbedder() {
...
+
+        protected abstract void process(ObjectNode record, JsonNode response);
+
         @Override
-        public void addEmbedRequests(SSCBulkRequestBuilder builder, UnirestInstance unirest, JsonNode record) {
+        public final void addEmbedRequests(SSCBulkRequestBuilder builder, UnirestInstance unirest, JsonNode record) {
             var id = record.get("id").asText();
-            builder.request(getBaseRequest(unirest)
-                    .routeParam("id", id)
-                    .queryString("limit", "-1"),
```
rsenden, line 47: I can see why this was removed here, as `issueDetails` endpoint doesn't support this query parameter. However, we now need to duplicate code to add this parameter to the other embedders, which is easy to forget in case we ever add other embedders. I think original code was better to ensure consistent behavior (always get all records if the endpoint supports the `limit` parameter), despite sending an unsupported query parameter on the `issueDetails` endpoint (which is simply ignored by SSC).

If we do want to avoid the latter, it would likely be better to somehow send `limit=-1` by default, unless an individual embedder chooses to opt out.

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: Added the changes

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueEmbedderSupplier.java
jmadhur87: Added the changes

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,85 @@
...
+    @Parameters(index = "0", arity = "1", descriptionKey = "fcli.fod.issue.get.vulnId")
+    private String vulnId;
+    @Mixin private FoDIssueEmbedMixin embedMixin;
+
+    @Override
+    protected IObjectNodeProducer getObjectNodeProducer(UnirestInstance unirest) {
+        FoDReleaseDescriptor releaseDescriptor = releaseResolver.getReleaseDescriptor(unirest);
+        String releaseId = releaseDescriptor.getReleaseId().toString();
+        JsonNode issue = findIssue(unirest, releaseDescriptor);
+        if ( issue instanceof ObjectNode issueObject ) {
```
rsenden: What if `issue` is not instance of `ObjectNode`? In that case, we don't add release id/name, don't pass the node through `transformRecord`, and output is undefined.

It would be better to have `findIssue` return an `ObjectNode`, either through simple cast (throwing an exception if it's not an `ObjectNode`, which I guess is very unlikely to happen), or explicit check (throwing an explicit exception in case the issue is not of type `ObjectNode`).

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,85 @@
...
+            issueObject.put("releaseId", releaseId);
```
rsenden: Doesn't the `issue` object already contain `releaseId` and `releaseName`?

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
```diff
@@ -0,0 +1,85 @@
...
+            issueObject.put("releaseName", releaseDescriptor.getReleaseName());
+            FoDIssueHelper.transformRecord(issueObject, IssueAggregationData.forSingleRelease(issueObject));
```
rsenden: I think this transformation is mostly meant for the `list` command, which can list issues from either an FoD app or release. Not sure whether it makes sense to add this data, although it does ensure that `list` and `get` commands output the same data. But if we want such consistency, then maybe the `get` command should accept either `--rel` or `--app`, with similar behavior as the `list` command (i.e., locate the given vuln id in either single release or all releases for given app, then transform based on appropriate `IssueAggregationData` factory method).

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: releaseId : already comes back from FoD. So put("releaseId", ...) just re-writes a value that's already there  it's redundant.
releaseName : does NOT come back from FoD. The list command also has to add it manually.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: Added the fix

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/issue/cli/cmd/FoDIssueGetCommand.java
jmadhur87: Right now get runs the same forSingleRelease aggregation transform that list uses, so both commands show the exact same columns. On a single-release get those aggregation fields (vulnIds, foundInReleases, ids, etc.) just end up as single-value entries.

So which do you prefer:

Keep it, get output stays idenical to list (consistent columns), even though the aggregation fields only ever hold one value here.
Drop it, get just returns the plain issue plus releaseName, no aggregation. Simpler, but get and list output will look a bit different.

PR #979, opened 2026-04-17
chore: Add `--fetch` option to `list` commands

rsenden:
```
feat: `fcli fod * list`: Add `--fetch` option on most `list` commands to fetch subset of records from FoD

feat: `fcli ssc * list`: Add `--fetch` option on most `list` commands to fetch subset of records from SSC

feat: `fcli sc-sast * list`: Add `--fetch` option on some `list` commands that utilize SSC REST endpoints to fetch subset of records from SSC

feat: `fcli sc-dast * list`: Add `--fetch` option on most `list` commands to fetch subset of records from ScanCentral DAST

feat: fcli `--style` option: Add `[no-]envelope` style for various output formats like JSON and YAML to allow for outputting paging and potentially other metadata

feat: fcli action framework: Emit `<key>.metadata` variable on `run.fcli` instructions to allow actions to access paging and potentially other metadata produced by the fcli command
```

### DIFF
--- a/fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/runner/processor/ActionStepProcessorRunFcli.java
+++ b/fcli-core/fcli-common/src/main/java/com/fortify/cli/common/action/runner/processor/ActionStepProcessorRunFcli.java
@@ -56,6 +56,7 @@ public class ActionStepProcessorRunFcli extends AbstractActionStepProcessorMapEn
     private static final String FMT_DEPENDENCY_SKIP_REASON = "%s.dependencySkipReason";
     private static final String FMT_EXIT_CODE = "%s.exitCode";
     private static final String FMT_RECORDS = "%s.records";
+    private static final String FMT_METADATA = "%s.metadata";
     private static final String FMT_STDOUT = "%s.stdout";
     private static final String FMT_STDERR = "%s.stderr";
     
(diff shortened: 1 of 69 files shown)


SangameshV:
Hello @rsenden 

Here are some of the review points given by Copilot

1. RequestObjectNodeProducer.applyAllFrom() does not call super.applyAllFrom(), and its applyFromObject helper has no case for IResponseMetadataCollector — so product helpers implementing it are never wired in for REST-based list commands.
2. assertNoExistingPagingParams in AbstractFetchRangeMixin.updateRequest() runs unconditionally on every request, even when --fetch is not specified.
3. ObjectNode[] metadataHolder = {null} single-element array in ActionStepProcessorRunFcli is an unconventional way to capture a mutable lambda variable.
4. FetchRangeConverter — input "1-2-3" is rejected due to a NumberFormatException on parseInt("2-3"), not explicit validation; the rejection is correct but the reason is implicit.
5. FetchRangeConverter has a redundant catch (FcliSimpleException e) { throw e; } block before the general catch (Exception e).
6. RecordWriterXml.doWriteEnvelopeEnd writes the metadata ObjectNode via writeTree, which serializes it as JSON-like structure inside XML rather than proper XML elements.
7. FetchRangeTest does not verify the error message text for any of the rejection cases, only that an exception is thrown.
8. FoDIssueListCommand — --fetch is silently accepted when --release is not specified, and only validated inside getObjectNodeProducer() at execution time rather than at parse time.

PR #991, opened 2026-05-04
feat: Implement suppression exclusions in tag mapping configuration

kireetivar:
This PR adds support for suppression exclusions in Aviator tag mapping.

Aviator can now skip auto-suppressing issues that match configured exclusions. The config shape is extensible, but for now we only support `categories`.

## Changes

- add `suppression_exclusions` to tag mapping config
- validate exclusion entries during YAML parsing
- apply exclusions when deciding whether to suppress issues
- update default YAML comments and native-image reflect config
- add tests for validation, YAML binding, matching, and suppression behavior

## Example

```yaml
suppression_exclusions:
  - categories:
      - "Privacy Violation"


rsenden (approved):
Given that there are no fcli command changes, the updated tag mapping file format has been agreed with Carl, and assuming Aviator team has tested this change, accepting this PR without detailed review.

PR #994, opened 2026-05-06
Adding new values to extensible list-type custom tags in SSC issue update during remediation

jmadhur87:
This PR enhances the SSC issue update command used during remediation to support adding new values to custom tags of type list when the tag is marked as extensible=true using new option --extend.
`ssc issue update --extend`

While verifying for ssc tag update command, found that value updates for non-LIST custom tags (TEXT, DECIMAL, DATE) were incorrectly reported as Updated in fcli, even though SSC made no change.
Added explicit validation and a clear error message so these unsupported value operations now fail fast instead of showing a false success.

Fix: Support adding new values to extensible list-type custom tags in SSC issue update during remediation([#959](https://github.com/fortify/fcli/issues/959) )


fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/custom_tag/helper/SSCCustomTagDefinitionHelper.java
```diff
@@ -0,0 +1,255 @@
...
+            body.set("valueList", buildUpdatedValueList(body, values, addValues, rmValues));
+        }
+        return body;
+    }
+
+    public int addValueToListTag(String tagGuid, String newValue) {
+        SSCCustomTagDescriptor desc = getDescriptorByCustomTagSpec(tagGuid, true);
+        ObjectNode body = (ObjectNode) desc.asJsonNode().deepCopy();
+        LinkedHashMap<String, ObjectNode> valueMap = buildValueMap(body);
+        if (!valueMap.containsKey(newValue)) {
```
SangameshV: A case mismatch in the keys can create duplicate list values in SSC. The `buildValueMap` method returns the `listValues` map with lowercase keys, and the `containsKey` check always returns `false` due to the case difference, thereby allowing duplicate list values to be added to the overall list.

**Note:** The same bug would cascade into `extendTagWithValue`, which caches the returned `newLookupIndex` in `tagInfo.getValueList()` a wrong index would be reused for all subsequent issues in the same command run.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
```diff
@@ -144,23 +145,38 @@ private String processDateValue(String value, String tagName) {
...
                 .collect(Collectors.joining(", "));
-        
-        throw new FcliSimpleException("Invalid value '" + value + "' for list custom tag '" + tagName + "'. " +
-                "Valid values are: " + validValues);
+        throw new FcliSimpleException("Invalid value '" + value + "' for custom tag '" + tagName + "'."
+                + " Supported values: " + validValues + "." + hint);
+    }
+    
+    private int extendTagWithValue(CustomTagInfo tagInfo, String newValue) {
+        int newIndex = new SSCCustomTagDefinitionHelper(unirest).addValueToListTag(tagInfo.getGuid(), newValue);
```
SangameshV, line 260: The `addValueToListTag` makes a redundant full **GET /customTags** call.

`extendTagWithValue` in `SSCIssueCustomTagHelper` creates a new `SSCCustomTagDefinitionHelper` instance that triggers lazy-loading of all custom tags via **GET /api/v1/customTags?limit=-1** just to get the descriptor for one tag. A targeted **GET /api/v1/customTags/{id}** would be more efficient.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueUpdateCommand.java
```diff
@@ -56,6 +56,8 @@ public class SSCIssueUpdateCommand extends AbstractSSCJsonNodeOutputCommand impl
     private String comment;
     @Option(names = {"--assign-user"})
     private String assignUser;
+    @Option(names = {"--extend"}, defaultValue = "false")
```
SangameshV, line 59: `--extend` at issue update level, not per-tag. 
If updating multiple custom tags in one command (e.g. --custom-tags TagA=NewVal,TagB=ExistingVal), which custom tag the `--extend` applies to? What if a custom tag is not extensible?
For the extensible custom-tag it should add a value and for the not extensible tags, proper error message must be displayed.

SangameshV (changes requested):
Also please check whether the `SSCCustomTagDescriptor` requires `extensible` field

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
```diff
@@ -57,10 +59,73 @@ public List<SSCIssueCustomTagAuditValue> processCustomTags(Map<String,String> cu
...
+            case TEXT:
+                return;
+            case DECIMAL:
+                if (!isUnset) {
+                    validateDecimalValue(tagName, tagValue);
+                }
+                return;
+            case DATE:
+                if (!isUnset) {
+                    processDateValue(tagValue, tagName);
```
rsenden: For other cases, we call a `validate` function, whereas here we call a `process` function, which looks weird, and we're not using the return value of the `processDateValue` method. Instead of explicitly validating the value, we're implicitly assuming that the `processDateValue` performs validation (which it does, but this is a code smell).

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
```diff
@@ -57,10 +59,73 @@ public List<SSCIssueCustomTagAuditValue> processCustomTags(Map<String,String> cu
...
+                if (value.equalsIgnoreCase(item.getLookupValue())) {
+                    return;
+                }
+            }
+        }
+        if (tagInfo.isExtensible() && extend) {
+            return;
+        }
+        String hint = tagInfo.isExtensible()
+                ? " Use --extend to add new values."
```
rsenden: This is a generic helper class that isn't aware of which fcli command it is being called from, and whether that command provides a `--extend` option. Maybe in the future, we call these helper methods from some other command that doesn't provide `--extend` option, thus showing an incorrect error message to the user.

Instead of passing a simple boolean `extend`, maybe better to introduce a small helper interface that provides a method like `checkExtendAllowed(...)` (or whatever else fits code structure); the command class would then provide an implementation that throws an appropriate error showing the actual option name.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
```diff
@@ -57,10 +59,73 @@ public List<SSCIssueCustomTagAuditValue> processCustomTags(Map<String,String> cu
...
     }
+
+    private void validateCustomTagValue(String tagName, String tagValue, Map<String, CustomTagInfo> tagInfoMap, boolean extend) {
+        CustomTagInfo tagInfo = tagInfoMap.get(tagName.toLowerCase());
+        if (tagInfo == null) {
+            throw new FcliSimpleException("Custom tag '" + tagName + "' is not available for this application version");
+        }
+
+        boolean isUnset = tagValue == null || tagValue.isBlank();
+        switch (tagInfo.getValueType()) {
```
rsenden, line 102: Can't we just skip validation altogether if `isUnset` is true, instead of repeating if-statement in each `case` block?

rsenden (changes requested):

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/custom_tag/helper/SSCCustomTagDefinitionHelper.java
jmadhur87: fixed using case-insensitive comparisons throughout.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
jmadhur87: fix: now uses a targeted single-tag fetch instead.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/cli/cmd/SSCIssueUpdateCommand.java
jmadhur87: each tag is checked individually; non-extensible tags throw a clear error message.
If any tag fails validation (e.g., non-extensible tag with new value), an exception is thrown immediately nothing gets updated

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
```diff
@@ -144,23 +209,38 @@ private String processDateValue(String value, String tagName) {
...
+                + " Supported values: " + validValues + "." + hint);
+    }
+    
+    private int extendTagWithValue(CustomTagInfo tagInfo, String newValue) {
+        int newIndex = new SSCCustomTagDefinitionHelper(unirest).addValueToListTagById(tagInfo.getId(), newValue);
+        ValueListItem newItem = new ValueListItem();
+        newItem.setLookupIndex(newIndex);
+        newItem.setLookupValue(newValue);
+        tagInfo.getValueList().add(newItem);
+        return newIndex;
```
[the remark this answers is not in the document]
jmadhur87: The current implementation already calls addValueToListTag(tagInfo.getGuid(), newValue), not addValueToListTagById(tagInfo.getId(), ...). The id field is never populated on CustomTagInfo and is not used in this code path. The scenario described in this comment does not apply to the actual implementation — no change needed.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/custom_tag/helper/SSCCustomTagDefinitionHelper.java
```diff
@@ -0,0 +1,262 @@
...
+    public int addValueToListTag(String tagGuid, String newValue) {
+        SSCCustomTagDescriptor desc = getDescriptorByCustomTagSpec(tagGuid, true);
+        return addValueToListTagInternal(desc.getId(), (ObjectNode) desc.asJsonNode().deepCopy(), newValue);
+    }
+
+    public int addValueToListTagById(String tagNumericId, String newValue) {
+        JsonNode response = unirest.get(SSCUrls.CUSTOM_TAG(tagNumericId)).asObject(JsonNode.class).getBody();
+        ObjectNode body = (ObjectNode) response.get("data").deepCopy();
+        return addValueToListTagInternal(tagNumericId, body, newValue);
+    }
```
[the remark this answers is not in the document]
jmadhur87: Added the changes

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
jmadhur87: Fixed by extracting [validateDateFormat()](vscode-file://vscode-app/c:/Users/mjain6/AppData/Local/Programs/Microsoft%20VS%20Code/1b50d58d73/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and calling it explicitly in the validation phase.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
jmadhur87: Fixed by introducing an [ExtendPolicy](vscode-file://vscode-app/c:/Users/mjain6/AppData/Local/Programs/Microsoft%20VS%20Code/1b50d58d73/resources/app/out/vs/code/electron-browser/workbench/workbench.html) interface. The command now provides the implementation that throws the error with the actual option name.

fcli-core/fcli-ssc/src/main/java/com/fortify/cli/ssc/issue/helper/SSCIssueCustomTagHelper.java
jmadhur87: Fixed by adding an early if (isUnset) return; guard before the switch, removing per-case repetition.

PR #1015, opened 2026-05-22
`fod sast-scan start`: Add `--in-progress-action` and `--entitlement-preference` options

gilseara:
## Summary

- Adds `--in-progress-action` and `--entitlement-preference` options to `fcli fod sast-scan start`, exposing the [`start-scan-advanced`](https://api.emea.fortify.com/swagger/ui/index#!/StaticScans/StaticScansV3_StartScanAdvanced) endpoint's `inProgressScanActionType` parameter (`DoNotStartScan` | `CancelScanInProgress` | `Queue`) and `entitlementPreferenceType` parameter.
- **Backward compatible**: when neither new option is passed, the command continues to use `start-scan-with-defaults` (no behavior change for existing callers, including the bundled `fod ci` action).
- When the advanced path is taken and `--in-progress-action` is not explicitly set, fcli defaults it to `Queue` instead of FoD's `DoNotStartScan`.
- Sends the API-correct value `CancelInProgressScan` over the wire (FoD's `start-scan-advanced` rejects `CancelScanInProgress`); the existing shared `InProgressScanActionType` enum keeps its current name so DAST behavior is unaffected — translation happens only at the SAST command boundary.

## Test plan

- [ ] `fcli fod sast-scan start --rel <app:rel> -f <pkg.zip>` — verify it still hits `start-scan-with-defaults` (no regression).
- [ ] `fcli fod sast-scan start --rel <app:rel> -f <pkg.zip> --in-progress-action=Queue` — verify request goes to `start-scan-advanced` with `inProgressScanActionType=Queue`.
- [ ] `fcli fod sast-scan start ... --in-progress-action=CancelScanInProgress` — verify the API receives `CancelInProgressScan` and the scan-in-progress is cancelled.
- [ ] `fcli fod sast-scan start ... --in-progress-action=DoNotStartScan` — verify FoD rejects/skips when a scan is already running.
- [ ] `fcli fod sast-scan start ... --entitlement-preference=SubscriptionOnly` — verify the entitlement preference is honored and `--in-progress-action` silently defaults to `Queue`.
- [ ] `fcli fod sast-scan start --help` — verify both new options show with the correct help text and completion candidates.
- [ ] `fcli fod dast-scan start --help` — confirm DAST still shows `CancelScanInProgress` as a valid value (unchanged).

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/scan/helper/sast/FoDScanSastHelper.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/scan/helper/sast/FoDScanSastHelper.java
@@ -67,7 +67,7 @@ public class FoDScanSastHelper extends FoDScanHelper {
                 .queryString("remdiationScanPreferenceType", (req.getRemdiationScanPreferenceType() != null ?
                         FoDEnums.RemediationScanPreferenceType.valueOf(req.getRemdiationScanPreferenceType()) : FoDEnums.RemediationScanPreferenceType.NonRemediationScanOnly))
                 .queryString("inProgressScanActionType", (req.getInProgressScanActionType() != null ?
-                        FoDEnums.InProgressScanActionType.valueOf(req.getInProgressScanActionType()) : FoDEnums.InProgressScanActionType.DoNotStartScan))
+                        req.getInProgressScanActionType() : FoDEnums.InProgressScanActionType.DoNotStartScan.toString()))
                 .queryString("scanTool", req.getScanTool())
                 .queryString("scanToolVersion", req.getScanToolVersion())
                 .queryString("scanMethodType", req.getScanMethodType());
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/sast_scan/cli/cmd/FoDSastScanStartCommand.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/sast_scan/cli/cmd/FoDSastScanStartCommand.java
@@ -70,11 +70,14 @@ public class FoDSastScanStartCommand extends AbstractFoDScanStartCommand {
             if (useAdvanced) {
                 FoDEnums.InProgressScanActionType inProgressAction = inProgressScanActionType != null
                         ? inProgressScanActionType : FoDEnums.InProgressScanActionType.Queue;
+                // FoD's start-scan-advanced expects 'CancelInProgressScan' rather than the enum's 'CancelScanInProgress'
+                String inProgressApiValue = inProgressAction == FoDEnums.InProgressScanActionType.CancelScanInProgress
+                        ? "CancelInProgressScan" : inProgressAction.name();
                 FoDScanSastStartRequest startScanRequest = requestBuilder
                         .entitlementPreferenceType(entitlementPreferenceType != null ? entitlementPreferenceType.name() : null)
                         .purchaseEntitlement(false)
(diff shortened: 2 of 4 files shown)


rsenden:
@kadraman Can you please review this enhancement? For example, does it make sense to dynamically switch between `start-scan-advanced` and `start-scan-with-defaults`, or would it be better to just always use `start-scan-advanced`? Probably good to also have Copilot do a review, but first would like your feedback on general approach.

kadraman (changes requested):
The `in-progress-action DoNotStarScan` is exiting with `FcliException` if a scan is in progress - this is the API response and not very good - but shouldn't we intercept and display a nicer message here:

```
>java -jar .\build\libs\fcli.jar fod sast-scan start -f ..\fortify-demo-app\fortifypackage.zip --release "Fortify Demo App:klee2_dev" --in-progress-action DoNotStartScan
FcliSimpleException: Error uploading file
        at com.fortify.cli.fod._common.rest.helper.FoDFileTransferHelper.uploadChunked(FoDFileTransferHelper.java:101)
Caused by: com.fortify.cli.common.rest.unirest.UnexpectedHttpResponseException: 
Request: POST https://api.emea.fortify.com/api/v3/releases/191706/static-scans/start-scan-advanced?entitlementPreferenceType=SubscriptionFirstThenSingleScan&purchaseEntitlement=false&remdiationScanPreferenceType=NonRemediationScanOnly&inProgressScanActionType=DoNotStartScan&scanTool=fcli&scanToolVersion=0.20260610.141015&scanMethodType=Other&fragNo=0&offset=0:   
  Reason: HTTP 422 Unprocessable Entity
  Body:   
    {"errors":[{"errorCode":2001,"message":"Cannot start scan as another scan is in progress"}]}
  ----
        at com.fortify.cli.common.rest.unirest.config.UnirestUnexpectedHttpResponseConfigurer$UnexpectedHttpResponseInterceptor.onResponse(UnirestUnexpectedHttpResponseConfigurer.java:36)
        at kong.unirest.CompoundInterceptor.lambda$onResponse$1(CompoundInterceptor.java:48)
```

gilseara:
Pushed updates for all the feedback:

**@kadraman — friendlier message when a scan's already running**
`fod sast-scan start` now catches FoD's HTTP 422 (`errorCode 2001`) and prints something readable instead of dumping the upload/HTTP stack trace:

```
Cannot start scan: another scan is already in progress for this release.
Use '--in-progress-action=Queue' to queue this scan, or
'--in-progress-action=CancelScanInProgress' to cancel the running scan and start a new one.
```

It walks the cause chain for an `UnexpectedHttpResponseException` with status 422 and the "another scan is in progress" reason, so anything else still bubbles up as before.

**Copilot**
- *Possible NPE on `remediationScanType`* — put the `remediationScanType != null` guard back before calling `getRemediationScanPreferenceType()`, like the original code and the other scan-start commands do.
- *Missing tests* — added three cases to `FoDScanSpec`:
  - `--help` shows both new options and all the completion candidates (`DoNotStartScan` / `CancelScanInProgress` / `Queue`)
  - `--in-progress-action=DoNotStartScan` while a scan's running → checks the new error message
  - `--in-progress-action=Queue` → hits the `start-scan-advanced` path and checks the scan starts

**@rsenden — always `start-scan-advanced` vs. the dynamic switch**
Kept the dynamic switch for now: if neither `--in-progress-action` nor `--entitlement-preference` is passed, we stay on `start-scan-with-defaults` so existing behavior (incl. the bundled `fod ci` action) doesn't change. Happy to move to always-advanced as a follow-up if you'd rather go that way — just let me know.

PR #1019, opened 2026-05-29
fix: fcli aviator ssc audit: Reject --skip-if-exceeding-quota with --…

kireetivar:
This change prevents users from combining `--skip-if-exceeding-quota` and `--folder-priority-order` on `fcli aviator ssc audit` These options represent conflicting behaviors, so the command now fails fast with a clear validation error instead of skipping audit execution before folder prioritization can apply.

The change also keeps the SSC bulk audit action aligned with the same validation and adds regression coverage for the direct command path.


fcli-core/fcli-app/src/test/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommandTest.java
```diff
@@ -0,0 +1,45 @@
...
+package com.fortify.cli.aviator.ssc.cli.cmd;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import com.fortify.cli.app._main.cli.cmd.FCLIRootCommands;
+
+import picocli.CommandLine;
+
+public class AviatorSSCAuditCommandTest {
```
rsenden: The `fcli-app` module currently only contains unit tests for asserting generic fcli characteristics, no module/command-specific tests. I think it would make more sense to move this test scenario to fcli-functional-test, using existing functional test infrastructure to expect an error when running this fcli command with invalid combination of options. Potentially, this would also allow you to test the same scenario on the bulkaudit.yaml action.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCAuditCommand.java
```diff
@@ -75,16 +76,19 @@ public class AviatorSSCAuditCommand extends AbstractSSCJsonNodeOutputCommand imp
     @Option(names = {"--tag-mapping"}) private String tagMapping;
     @Option(names = {"--no-filterset"}) private boolean noFilterSet;
     @Option(names = {"--folder"}, split = ",") @DisableTest(DisableTest.TestType.MULTI_OPT_PLURAL_NAME) private List<String> folderNames;
-    @Option(names = {"--skip-if-exceeding-quota"}) private boolean skipIfExceedingQuota;
+    @ArgGroup(exclusive = true, multiplicity = "0..1") private QuotaHandlingArgGroup quotaHandlingArgGroup = new QuotaHandlingArgGroup();
```
rsenden, line 79: Theoretically, this should be considered a breaking change as previous fcli versions allowed both options to be specified on a single invocation. If anyone has both options in some workflow/automation, this would now result in an error being reported instead of silently ignoring one of the options. As previous behavior was incorrect, I guess it's fine to classify this as a bug fix only though, not a breaking change.

rsenden (changes requested):

rsenden (approved):

PR #1035, opened 2026-06-15
Add Scan Central Dast attribute in ssc attribute list definition command

jmadhur87:
This change adds support for retrieving ScanCentral DAST attributes from SSC by including the required `extraCategories` parameter in the attributes API request. As a result, `ssc attr lsd` now lists ScanCentral DAST attributes, and applications created using these attributes are populated correctly.

fix: Add Scan Central Dast attribute in ssc attribute list definition command


rsenden:
@jmadhur87, as the PR is still in draft, I assume you're still working on this? I think ScanCentral DAST attributes on SSC were recently introduced; please check whether this change causes any issues on older SSC versions, i.e., whether older SSC versions just ignore this query parameter, or throw an error.

jmadhur87:
@rsenden, tested this by adding an unknown query parameter to the SSC API request. SSC ignored the parameter and processed the request successfully without returning an error, so older SSC versions should not be impacted by this change.

PR #1037, opened 2026-06-17
feat(ci): add Aviator remediation action with automated PR creation (SSC & FoD)

SangameshV:
## Summary

Introduces a new FCLI action to automatically apply Aviator remediation fixes and create a pull request with the generated changes.

This implementation supports both **SSC (Software Security Center)** and **FoD (Fortify on Demand)** pipelines.
## Background

Aviator generates remediation changes for identified issues, but there was no automated way to:

- Apply fixes to the repository
- Commit updates
- Push changes to a new branch
- Create a pull request for review

### ✅ New Feature: Automated Remediation Action

- Implemented FCLI action to:
  - Apply Aviator-generated remediations
  - Stage and commit updated files
  - Create a remediation branch
  - Push the branch to the remote repository
  - Open a pull request against the base branch

---


fcli-core/fcli-common-action/src/main/java/com/fortify/cli/common/action/helper/git/ActionGitSpelFunctions.java
```diff
@@ -0,0 +1,486 @@
...
+ * staging files, committing, and pushing changes to a remote.
+ *
+ * Available via the {@code #git} SpEL variable in action YAML files.
+ *
+ * @author Sangamesh Vijayakumar
+ */
+@Reflectable
+@SpelFunctionPrefix("git.")
+@Slf4j
+public class ActionGitSpelFunctions {
```
rsenden, line 73: Some of the methods in this class are very long; please refactor using private helper methods. Also, please make sure that proper imports are used instead of fully qualified class names.

fcli-core/fcli-common-action/src/main/java/com/fortify/cli/common/action/helper/git/ActionGitSpelFunctions.java
```diff
@@ -0,0 +1,486 @@
...
+            if (!dir.exists()) { return null; }
+            var builder = new FileRepositoryBuilder().findGitDir(dir);
+            if (builder.getGitDir() == null) { return null; }
+            return new Git(builder.build());
+        } catch (Exception e) {
+            return null;
+        }
+    }
+
+    private CredentialsProvider detectCredentialsProvider() {
```
rsenden: This tightly binds generic Git functionality to individual CI systems. When we add support for a CI system, we'd expect having to create corresponding CI-specific classes, but it may be easily overlooked that this method needs to be updated as well.

Can we refactor this, for example by having some `getGitCredentialsProvider` method in each of the CI implementations, and then either have the fcli action yaml explicitly retrieve credentials from currently active CI environment and pass this to ActionGitSpelFunctions methods, or have ActionGitSpelFunctions somehow retrieve this from the currently active CI environment implementation?

Please check what architecture would fit best here, for example in which of the CI-specific classes the `getGitCredentialsProvider` method should live (if we choose to have such a method), and how ActionGitSpelFunctions can gain access to that. 

Also, what if for example workflow is running in ADO, but repo is on GitHub? I guess in that case, `SYSTEM_ACCESSTOKEN` wouldn't give access to the GitHub repo, so would be good to do some research as to how other (non-Fortify) pipeline integrations handle such scenarios. For now, we can consider this an edge case, but it might affect optimal architecture for the above, so good to have a look at this.

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/apply-remediations.yaml
```diff
@@ -0,0 +1,39 @@
...
+author: Fortify
+usage:
+  header: Apply Aviator remediations
+  description: |
+    This action applies Aviator auto-remediation fixes to source code for a given
+    FoD release. It downloads the audited FPR with remediation suggestions and applies
+    the fixes to the local source tree.
+    
+    This action requires:
+    - A FoD session to be active (for downloading the audited FPR with remediations)
```
rsenden: This is true for all actions in the FoD module, so unnecessary to explicitly list this as requirement

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/ci.yaml
```diff
@@ -135,6 +135,20 @@ steps:
         on.success:
           - var.set: { postScan.skipReason: } # Reset postScan.skipReason to allow post-scan tasks to run
              
+      APPLY_REMEDIATIONS:
```
rsenden: For consistency with other Aviator-related steps, probably better to rename steps and env vars to something like `AVIATOR_REMEDIATE`.

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/ci.yaml
```diff
@@ -195,15 +195,32 @@ steps:
               skip.if-reason:
                 - ${postScan.skipReason}                               # Skip if no scans were run
                 - ${aviator.skipReason}                                # Skip if Aviator audit is to be skipped
-                
-                  
+              on.success:
+                - var.set:
```
rsenden, line 199: Why do we need to explicitly retrieve artifact id from the output, rather than just relying on fcli `--store` option and using `::aviator_audit::` (like we did before) instead of `${aviator.artifactId}`

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/ci.yaml
```diff
@@ -135,6 +135,20 @@ steps:
...
+        cmd: ${#actionCmd('APPLY_REMEDIATIONS', 'fod', 'apply-remediations')} "--progress=none"
+        skip.if-reason:
+          - ${#actionCmdSkipFromEnvReason('APPLY_REMEDIATIONS', true)} # Skip unless DO_APPLY_REMEDIATIONS==true or APPLY_REMEDIATIONS_ACTION/EXTRA_OPTS defined
+          - ${postScan.skipReason}                                     # Skip if no scans were run
+          - ${#env('DO_AVIATOR_AUDIT')!='true'?'Aviator audit not enabled (DO_AVIATOR_AUDIT!=true), no remediations available':''} # Skip if Aviator audit was not enabled
```
rsenden, line 144: Any more reliable way to check whether Aviator Audit was run? Maybe check FoD whether Aviator audit was run, or, even better, check whether remediations are available?

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/ci.yaml
```diff
@@ -135,6 +135,20 @@ steps:
...
+          - ${SAST_WAIT.dependencySkipReason}                          # Skip if SAST scan/wait was skipped or failed
+
+      CREATE_PR:
```
rsenden: In a CI environment, it doesn't make sense to apply remediations but not raise a PR; if Aviator remediation was performed, we should unconditionally raise a PR (i.e., `DO_AVIATOR_REMEDIATE` controls both applying remediations and raising the associated PR).

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/create-pr.yaml
```diff
@@ -0,0 +1,177 @@
+# yaml-language-server: $schema=https://fortify.github.io/fcli/schemas/action/fcli-action-schema-dev-2.x.json
```
rsenden: This file seems to be almost an exact duplicate of the FoD `create-pr.yaml` action (apart from `rest.target.default` value, but as we don't do any SSC/FoD REST calls, we don't need this); please avoid this duplication, for example:
- We still have code in `ActionLoaderHelper` to load actions from `commonActionsResourceZip`, although I don't know whether we still have Gradle build step to generate `com/fortify/cli/common/actions.zip`, and whether that requires common actions to be in a specific module (`fcli-common-action` would now be the most appropriate module for defining common actions). If we'd move `create-pr.yaml` to a common action, it would automatically be available for `fcli fod action`, `fcli ssc action`, and `fcli action`.
- Simply move to `fcli-action`, i.e., we'd run `fcli action run pr-comment` for both SSC and FoD workflows
- Get rid of this action altogether (see other comments about architecture)

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/apply-remediations.yaml
```diff
@@ -0,0 +1,39 @@
...
+      rel: ${#env('FOD_RELEASE')?:global.ci.defaultFortifyRepo}
+  - if: ${rel==null}
+    throw: FoD release must be specified through FOD_RELEASE environment variable
+
+  - var.set:
+      sourceDir: ${global.ci.sourceDir?:'.'}
+
+  - run.fcli:
+      APPLY_REMEDIATIONS:
+        cmd: fod aviator apply-remediations --rel "${rel}" --source-dir "${sourceDir}"
```
rsenden: For interactive use, having this action doesn't make much sense; none of the CI env detection and FoD/SSC env vars are relevant in non-CI environments, hence users can just as well run the `apply-remediations` command directly. For CI use, we already have all the relevant data available in `ci.yaml`, so it would make more sense to directly run `apply-remediations` command from `ci.yaml`.

fcli-core/fcli-common-action/src/main/java/com/fortify/cli/common/action/helper/git/ActionGitSpelFunctions.java
```diff
@@ -0,0 +1,486 @@
...
+
+    @SpelFunction(cat=util, desc="Stages all modified and new files in the working tree for commit.",
+            returns="`true` if files were staged successfully")
+    public boolean addAll(
+            @SpelFunctionParam(name="sourceDir", desc="directory inside a git working tree") String sourceDir) {
+        try (var git = openGit(sourceDir)) {
+            if (git == null) {
+                throw new FcliSimpleException("Not a git repository: " + sourceDir);
+            }
+            git.add().setUpdate(true).addFilepattern(".").call();
```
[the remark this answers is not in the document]
SangameshV: The current implementation of addAll() using setUpdate(true) is intentional. It ensures that only modified or deleted files are staged, while excluding newly created files. This was done to prevent unrelated files generated during CI action runs from being accidentally committed.
This behavior was explicitly introduced as part of [commit 9c7a5bc](https://github.com/fortify/fcli/pull/1037/changes/9c7a5bc7a9ff99eed8e37674d30c676c8128b454) to limit commits only to actual remediation changes.

Refer to the below screenshot showing previously committed unwanted files.
image

fcli-core/fcli-common-action/src/main/java/com/fortify/cli/common/action/helper/git/ActionGitSpelFunctions.java
```diff
@@ -0,0 +1,486 @@
...
+        if (StringUtils.isBlank(sourceDir)) { return null; }
+        var dir = Path.of(sourceDir).toAbsolutePath().normalize().toFile();
+        if (!dir.exists()) { return null; }
+        FileRepositoryBuilder builder = new FileRepositoryBuilder().findGitDir(dir);
+        if (builder.getGitDir() == null) { return null; }
+        try (Repository repo = builder.build()) {
+            var mapper = JsonHelper.getObjectMapper();
+            var remote = selectRemote(repo);
+            var remoteUrl = remote == null ? "origin" : repo.getConfig().getString("remote", remote, "url");
+            var names = deriveRepoNames(dir.getName(), remoteUrl);
```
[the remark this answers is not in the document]
SangameshV: Correct!! Even the suggested line/s of code still relies on unsafe `origin` fallback. I’ll throw an exception if the remote URL is null or blank as below.
```var remote = selectRemote(repo);
   String remoteUrl = null;
   var remoteUrl = remote != null
           ? repo.getConfig().getString("remote", remote, "url")
           : null;
   if (remoteUrl == null || remoteUrl.isBlank()) {
        throw new IllegalStateException("No valid remote URL configured for repository");
   }

**Review by rsenden — CHANGES_REQUESTED** [review]
Note that some of the comments posted on FoD files may also apply to SSC, or vice versa.

Other than individual comments, I'm wondering whether we can improve architecture to be more logical and more user-friendly, which would supersede most of the individual comments.

Please investigate whether something like the following in the `ci.yaml` files would work:

```
AVIATOR_REMEDIATE:
  cmd: ${#fcliCmd(<SSC/FoD apply-remediations cmd>)}
  skip.if-reason:
    <same as before, possibly enhanced for FoD to check remediations are available>
  on.success:
    - var.set:
         createPrResult: ${#<CI-specific SpEL function>(subject, body, ...)}
    - log.info: Aviator remediations ${global.ci.prTerminology} created: ${createPrResult.prLink}
```

I.e.:
- We'd get rid of `apply-remedations.yaml` and `create-pr.yaml` actions
- Functionality from `create-pr.yaml` would move to Java code, exposed through SpEL function(s)
- Running apply-remediations and creating the PR/MR consolidated in a single `ci.yaml` step

Main questions:
- How to easily call a CI-specific SpEL function, like `#github.repo().raisePR(...)` vs `#gitlab.project().raiseMR(...)` in a generic way, without hardcoding knowledge about supported CI systems and what functions they expose, directly in `ci.yaml`. Maybe just add a `raisePR(...)` method on `ActionCiSpelFunctions` (or have a separate SpelFunctions class that handles cross-CI PR logic), which calls `detect()` and then calls the appropriate CI-specific SpEL function on the detected CI system?
- How to handle add/commit/push? Probably easiest to do this from Java code, such that SSC/FoD `ci.yaml` files just need to call a single SpEL function

Also something to consider; what if the build that was executed by `scancentral package` caused any changes in current git state (like adding generated code in non-ignored paths). We'd want the PR to only include code that was added/removed/updated by the `apply-remediations` command, so possibly we should stash any changes before running apply-remediations, and unstash after pushing the Aviator changes/raising the PR?

fcli-core/fcli-common-action/src/main/java/com/fortify/cli/common/action/helper/git/ActionGitSpelFunctions.java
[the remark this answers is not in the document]
rsenden: Although I think current Aviator remediation only changes existing files, what if a future version can also create new files, for example adding some validator class that's than called from updated code. If we don't include new files, the PR would contain broken code.

In my manual comments, I added a suggestion to use git stash/unstash, although given your screenshot, that might not work well either; no idea what would happen if you stash/unstash the fcli.log file that current fcli invocation is actively writing to.

I don't mind keeping it simple for now with just including changed files, but we should at least add a comment explaining this; why we don't add new files (build output that shouldn't be committed, fcli state data, ...), and how this can causes issues if future Aviator versions do create new files.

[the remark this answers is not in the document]
SangameshV: Adding `fcli.log` to `.gitignore` makes sense and avoids committing that file. However, using `git stash`/`unstash` in CI introduces complexity and potential inconsistencies, especially with untracked files and conflict handling.

For now, it’s safer to keep the current approach of **committing only modified files** and explicitly document this behavior. We can revisit this if future Aviator versions require inclusion of newly created files, potentially with a controlled/whitelisted inclusion strategy instead of relying on stash.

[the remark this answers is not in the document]
rsenden: Although I understand the use case, the method name `addAll` doesn't properly reflect actual functionality (which only adds changed files, doesn't add new files). At least method name should be updated to reflect actual behavior, or possibly do some refactoring to make this configurable by the caller.

rsenden (changes requested):
To start with, note that there have been some general changes/improvements to CI-related environment classes/SpEL functions in `dev/v3.x`, which might (or might not) be useful for improving the code in this PR.

I have some doubts about current implementation:
- Looks like most of the earlier review comments haven't yet been addressed, like long methods, names of steps in `ci.yaml`, CI token handling, proper separation between generic git-related code and CI-specific credential providers, ... Note though that implementation might change based on the below, so best to handle those first.
- `ci` action usually calls a single sub-action/fcli command to perform a specific task, but for this use case, it now needs to call three separate cmds/actions; first apply-remediations, then git-push-changes, then create-pr.
- Apparently, JGit doesn't natively integrate with CI-provided credentials, and according to AI, there are some other edge cases that require special attention in JGit compared to plain `git` invocations from CI pipelines. Also, looks like quite a lot of code is needed to accomplish tasks that would only require simple a `git` command if we'd do this from command line.

Please research feasibility of the following:
- Only have a single `AVIATOR_REMEDIATE` step in `ci.yaml`, without separate steps for pushing changes and raising PR. Some potential ideas:
    - Have a single `raise-remediation-pr` action in `fcli-action` that runs either SSC or FoD apply-remediations command (select based on CLI arg), commits/pushes the changes, then raises PR (on supported CI systems)
    - Have two `raise-remediation-pr` actions; one in fcli-fod, one in fcli-ssc; same as above, but automatically calling the appropriate apply-remediations cmd
    - Have CI-specific actions like `gitlab-remediations-mr` and `github-remediations-pr`; either single copy in `fcli-action` or two copies in `fcli-fod` and `fcli-ssc`. Advantages: can use CI-specific terminology (MR vs PR), users and `detect-env` can easily check availability for current CI system (and maybe fall back to a generic `push-remediations` action for CI systems on which we don't support raising PR yet), most in line with current action structure
    - If needed, to reduce duplication across multiple actions, have a special-purpose SpEL function that handles add+commit+push specifically for this use case
- Use `ProcessBuilder` for running plain `git add/commit/push commands`, instead of doing everything through JGit
    - Would likely greatly reduce amount of code in this PR; no need for remote/branch detection, no need for credentials handling, ...
    - Likely results in better CI integration / better user experience; `git` is usually properly configured for the checked out repo, whereas we'd need to manually handle this (including edge cases) with JGit.
    - In general, invoking external processes should be avoided, but warranted in this case given likely better CI integration, code simplification, and we can assume that `git` is available on CI systems & dev workstations
    - We can either provide wrapper SpEL functions (like `add`, `commit`, ...) for specific git operations (somewhat more secure, not exposing dangerous git cmds), or generic `#git.run(<args>)` SpEL function and/or `run.git` YAML instruction, somewhat similar to `run.fcli` (much more flexible)

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/apply-remediations.yaml
SangameshV: Handled to run the `apply-remediations` fcli command from `ci.yaml`

fcli-core/fcli-ssc/src/main/resources/com/fortify/cli/ssc/actions/zip/create-pr.yaml
SangameshV: A single generic fcli action `create-pr.yaml` to create a PR replaces these 2 fcli actions, which are almost similar/same.

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/ci.yaml
SangameshV: Taken care to run create-pr step if there were remediations applied.

fcli-core/fcli-fod/src/main/resources/com/fortify/cli/fod/actions/zip/apply-remediations.yaml
SangameshV: Removed this file as part of the newer architecture changes.

rsenden (approved):
After latest refactoring, this is good enough for demo purposes; any bugs can be fixed later.

PR #1050, opened 2026-07-06
feat: Aviator apply-remediations issue-id filtering for SSC/FoD

kireetivar:
> Note: This PR will be merged into `feat/v3.x/aviator/26.4`

This PR adds remediations cache download/apply workflows for SSC and FoD so teams can download remediations once into a cache zip and apply them locally, including optional issue-level filtering.

What changed
- Added new cache download commands:
  - `fcli aviator ssc download-remediations-cache`
  - `fcli fod aviator download-remediations-cache`
- Extended existing apply commands:
  - `fcli aviator ssc apply-remediations`
  - `fcli fod aviator apply-remediations`
- Added `--from-cache` support to SSC and FoD apply-remediations.
- Added `--issue-ids` support for cache-based apply-remediations.
- Added shared cache zip handling for manifest, entry ordering, checksum validation, reading, and writing.
- Added SSC support for cache generation from:
  - `--artifact-id`
  - `--latest`
  - `--all`
- Added FoD support for cache generation from:
  - `--release` / `--rel`
- Added resilient handling for cache entries that do not contain `remediations.xml`.

Behavior
- With `--from-cache`:
  - Apply-remediations reads FPRs from the cache zip in manifest order.
  - SSC and FoD process cached entries sequentially.
- With `--issue-ids`:
  - Only remediations whose `instanceId` matches requested ids are considered.
  - `--issue-ids` is only supported together with `--from-cache`.
- Without `--issue-ids`:
  - Existing apply-remediations behavior is unchanged for non-cache flows.
- If a cached FPR does not contain `remediations.xml`:
  - The entry is skipped instead of failing the entire cache-based run.

SSC cache behavior
- `download-remediations-cache --all` stores artifacts in chronological order.
- `apply-remediations --from-cache` processes entries in the same stored order.
- Output reports:
  - processed entries
  - skipped entries
  - processed artifact ids

FoD cache behavior
- `download-remediations-cache` stores the selected release FPR in the cache zip.
- `apply-remediations --from-cache` processes entries in cache order.
- Output reports:
  - processed entries
  - processed release ids`

apply-remediations command additions
- `fcli aviator ssc apply-remediations --from-cache <zip>`
- `fcli aviator ssc apply-remediations --from-cache <zip> --issue-ids <id1,id2,...>`
- `fcli fod aviator apply-remediations --from-cache <zip>`
- `fcli fod aviator apply-remediations --from-cache <zip> --issue-ids <id1,id2,...>`


fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCDownloadRemediationsFprCommand.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Override
+    public String getActionCommandResult() {
+        return "REMEDIATIONS_FPR_DOWNLOADED";
+    }
+
+    @Override
+    public boolean isSingular() {
+        return true;
```
rsenden: What's the rationale for having singular output with artifact id's embedded in that single output record, rather than outputting separate records for every individual artifact? Unless there are good reasons against the latter, that's the more common fcli convention.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCDownloadRemediationsFprCommand.java
```diff
@@ -0,0 +1,155 @@
...
+                progressWriter);
+    }
+
+    private ObjectNode buildSingleResultNode(SSCArtifactDescriptor artifact, Path destination) {
+        ObjectNode result = artifact.asObjectNode();
+        result.put("artifactsDownloaded", 1);
+        result.putArray("artifactIds").add(artifact.getId());
+        result.putArray("files").add(destination.toString());
+        result.put("file", destination.toString());
+        result.put(IActionCommandResultSupplier.actionFieldName, getActionCommandResult());
```
rsenden: fcli output framework automatically sets the 'action' column based on `getActionCommandResult`; explicitly setting this field is only necessary if you want to display dynamic action values, i.e., if that value depends on execution flow/result.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCDownloadRemediationsFprCommand.java
```diff
@@ -0,0 +1,155 @@
...
+        ArrayNode files = result.putArray("files");
+        ArrayNode artifactIds = result.putArray("artifactIds");
+        for (int i = 0; i < artifacts.size(); i++) {
+            SSCArtifactDescriptor artifact = artifacts.get(i);
+            Path destination = outputDir.resolve(String.format("%03d_remediations_artifact_%s.fpr", i + 1, artifact.getId()));
+            downloadArtifact(unirest, artifact, destination, logger, progressWriter);
+            files.add(destination.toString());
+            artifactIds.add(artifact.getId());
+        }
+        result.put(IActionCommandResultSupplier.actionFieldName, getActionCommandResult());
```
rsenden: fcli output framework automatically sets the 'action' column based on `getActionCommandResult`; explicitly setting this field is only necessary if you want to display dynamic action values, i.e., if that value depends on execution flow/result.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCDownloadRemediationsFprCommand.java
```diff
@@ -0,0 +1,155 @@
...
+    private void validateDestinationOptions() {
+        if (artifactSelector.isAllOpenIssuesSelected()) {
+            if (outputFile != null) {
+                throw new FcliSimpleException("-f/--file cannot be used with --all; use --dir");
+            }
+            if (outputDir == null) {
+                throw new FcliSimpleException("--dir must be specified when using --all");
+            }
+        } else if (outputDir != null) {
+            throw new FcliSimpleException("--dir can only be used with --all; use -f/--file for a single FPR download");
```
rsenden: Why is `--dir` not supported for single-file downloads, using the default FPR file name as defined in `getSingleDestination`.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/mixin/AviatorSSCRemediationsFprDownloadSelectorMixin.java
```diff
@@ -0,0 +1,88 @@
...
+    @Getter
+    public static class ArtifactSelectionArgGroup {
+        @Option(names = {"--artifact-id"}, required = true, descriptionKey = "fcli.aviator.ssc.download-remediations-fpr.artifact-id")
+        private String artifactId;
+
+        @Option(names = {"--latest"}, required = true, descriptionKey = "fcli.aviator.ssc.download-remediations-fpr.latest")
+        private boolean latest;
+
+        @Option(names = {"--all"}, required = true, descriptionKey = "fcli.aviator.ssc.download-remediations-fpr.all")
+        private boolean allOpenIssues;
```
rsenden: Weird field name for 'all FPRs'

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/mixin/AviatorSSCApplyRemediationsArtifactSelectorMixin.java
```diff
@@ -53,6 +59,11 @@ public static class ArtifactSelectionArgGroup {
 
         @Option(names = {"--all"}, required = true, descriptionKey = "fcli.aviator.ssc.apply-remediations.all")
         private boolean allOpenIssues;
+
+        @Option(names = {"--fpr"}, required = true, arity = "1..*", paramLabel = "<file>",
+                descriptionKey = "fcli.aviator.ssc.apply-remediations.fpr")
+        @DisableTest({TestType.MULTI_OPT_SPLIT, TestType.MULTI_OPT_PLURAL_NAME, TestType.OPT_ARITY_VARIABLE})
```
rsenden: I guess the rationale for non-standard arity and the disabled tests is to allow this to work with wildcard patterns, i.e., `--fpr /path/to/downloads/*.fpr`? Although picocli does its best to handle multi-arity options, we encountered issues with this in the past, hence we usually disallow this.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCDownloadRemediationsFprCommand.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        result.put(IActionCommandResultSupplier.actionFieldName, getActionCommandResult());
+        return result;
+    }
+
+    private Path getSingleDestination(SSCArtifactDescriptor artifact) {
+        return outputFile == null ? Path.of(String.format("remediations_artifact_%s.fpr", artifact.getId())) : outputFile.toPath();
+    }
+
+    private void validateDestinationOptions() {
```
rsenden: In general, I don't like exclusive sets of options as it affects user experience (difficult to figure out which sets of options can be combined), nor manual validation instead of picocli handling option exclusivity (also to properly display this in command syntax). Any way around this, by reorganizing ArgGroups?

rsenden (changes requested):
I've done a partial review and added some individual code comments, which mostly originate from current (command) design. Can we improve the design, thereby automatically resolving most of these comments? 

I think the key issue is that these commands operate on individual FPR files, thus requiring separate `--file`/`--dir` output options depending on how many FPRs are being downloaded, and deviating from fcli arity conventions to pass these FPRs through wildcards to `apply-remediations` commands. Also, when depending on user-specified dirs/wildcards, there's a risk that incorrect FPR files get picked up, for example from a previous run, possibly even from a different app version/release.

Three alternative approaches come to mind (but maybe you can think of other approaches):
1. Get rid of `download-remediations-fpr` commands, and instead add caching options on `apply-remediations` commands, for example a `--[no-]use-cache` option that checks fcli state directory whether cached FPR files for current set of args (`--av`/`--rel` & artifact selection criteria) is available, downloading these to the cache on the fly if not yet available or outdated. Main complexity is cache management, for example requiring extra commands/options for updating or clearing cache, detecting outdated cache, ...
2. Have `download-remediations-fpr` (possibly renamed to something like `prepare-remediations-cache` output a single 'remediations cache' zip file; depending on FPR selection criteria, this zip-file can contain either single or multiple FPR files, which users can then pass to `apply-remediations` cmd using `--from-cache=<zip-file>`
3. Combination of the above; have `apply-remediations --use-cache=<zip-file>`; if given zip-file exists, it's used as is, otherwise it's created by downloading FPR files based on current FPR selection criteria. Possibly, the zip-file could contain metadata like app version/release and FPR selection criteria; if there's a mismatch with current selection criteria on the `apply-remediations` command, an error would be thrown.

kireetivar:
Thanks for the review comments. I’m planning to revise the PR away from the current raw-FPR workflow.

Proposed solution:

- I am more inclined toward your option 2, with one explicit cache/archive file produced by the download command and consumed by apply.
- Replace `download-remediations-fpr` with `download-remediations-cache`.
- Remove the multi-arity `--fpr` option completely.
- Use a single cache zip with `manifest.json` and ordered FPR entries.

Example `manifest.json`:

```json
{
  "schemaVersion": 1,
  "kind": "aviator-remediations-cache",
  "product": "ssc",
  "createdAt": "2026-07-16T12:00:00Z",
  "selection": {
    "mode": "all",
    "appVersionId": "10001",
    "since": "2026-07-01T00:00:00Z"
  },
  "entries": [
    {
      "order": 1,
      "artifactId": "12345",
      "uploadDate": "2026-07-10T08:00:00Z",
      "path": "fprs/001_artifact_12345.fpr",
      "sha256": "8f14e45fceea167a5a36dedd4bea2543"
    },
    {
      "order": 2,
      "artifactId": "67890",
      "uploadDate": "2026-07-11T09:30:00Z",
      "path": "fprs/002_artifact_67890.fpr",
      "sha256": "45c48cce2e2d7fbdea1afc51c7c6ad26"
    }
  ]
}
```

- Use `--from-cache <zip>` on `apply-remediations`, but not as a standalone source. The normal selection options are still required:
  - SSC: `--artifact-id`, or `--latest/--all --av ...`
  - FoD: `--rel`
- If the cache exists, apply validates the manifest against the current selection. If it does not exist, the user should create it first with `download-remediations-cache`.
- `--issue-ids` will require `--from-cache`, so filtered/partial applies do not repeatedly download from FoD/SSC.
- The manifest will include product, server, selection metadata, ordered entries, upload dates, and required `sha256`; apply will verify checksums before modifying source.
- Use the same `--from-cache` option name for SSC and FoD.

If you are ok with this proposal, I have a few questions I’d like your input on:

1. Are the proposed option/command names OK: `download-remediations-cache` and `--from-cache`?
2. Should `--file` be required for `download-remediations-cache`, or should we provide a deterministic default cache filename?
3. Should `download-remediations-cache` overwrite existing files by default, fail by default, or require an explicit `--overwrite` option?
4. If the selected SSC/FoD criteria resolve to zero matching artifacts, should we create an empty valid cache, or should cache download fail?

rsenden:
I don't have a strong preference for either option 2 or 3:
- Option 2 provides a more explicit split between creating the cache, and consuming the cache, but requires an extra command to be run
- Option 3 allows for simply running the same command (with different issue id's) multiple times, automatically creating and consuming the cache as needed, but it may be slightly confusing that the `--use-cache` is used for both initial creation of the cache, and consuming it in subsequent runs

For option 2, I think it's better to have `--from-cache` being mutually exclusive with `--av`/`--rel` and artifact selection options; i.e., users either apply remediations from cache, or from the existing artifact selection options. This is more user-friendly (no need to repeat same options on `download-remediations-cache` and `apply-remediations`), and saves us from having to validate that `apply-remediations` options match the cache.

For option 3, we would need to verify options against cache, but then need to decide what to do if there's a mismatch; throw an error, replace the cache based on current selection options, ...

> 1. Are the proposed option/command names OK: `download-remediations-cache` and `--from-cache`?

Yes, if we go for option 2

> 2. Should `--file` be required for `download-remediations-cache`, or should we provide a deterministic default cache filename?

I think requiring `--file` is better, providing a clear link between `download-remediations-cache` and `apply-remediations`

> 3. Should `download-remediations-cache` overwrite existing files by default, fail by default, or require an explicit `--overwrite` option?

Use `CommonOptionMixins::RequireConfirmation` mixin (see existing usages in fcli as example)

> 4. If the selected SSC/FoD criteria resolve to zero matching artifacts, should we create an empty valid cache, or should cache download fail?

Not sure; probably should have same behavior as current non-cache based `apply-remediations`. Output should probably show the number of downloaded artifacts.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheWriter.java
```diff
@@ -0,0 +1,337 @@
...
+    private RemediationsCacheWriter(Path destination, String product, Map<String, String> selection) {
+        this.destination = destination;
+        this.workPath = workPathFor(destination);
+        this.manifest = newManifest(product, selection);
+        FileSystem fs = null;
+        Exception error = null;
+        try {
+            // ZipFS on sibling work file; existing destination is left intact until publish.
+            fs = ZipHelper.createZipFileSystem(workPath);
+            Files.createDirectories(fs.getPath(RemediationsCacheConstants.FPRS_DIR));
```
rsenden: I think state management in this class can be significantly simplified if directory creation is moved outside of the constructor:
- Constructor just performs prerequisite ops/checks and then `this.zipFs = ZipHelper.createZipFileSystem(workPath)`
- Any exceptions in constructor can be immediately thrown; no need to store error and check/handle elsewhere
- No need to keep `closed` status, as `close()` would only need to be called by try-with-resources

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheWriter.java
```diff
@@ -0,0 +1,337 @@
...
+
+    /**
+     * Opens a work zip next to {@code destination} and prepares an empty manifest.
+     * On initialization failure, {@link #close()} runs before the exception is rethrown so
+     * cleanup is not split across factory and AutoCloseable paths.
+     */
+    public static RemediationsCacheWriter create(Path destination, String product, Map<String, String> selection) {
+        FcliSimpleException.throwIf(destination == null,
+                "-f/--file must specify a remediations cache zip path");
+        RemediationsCacheWriter writer = new RemediationsCacheWriter(destination, product, selection);
```
rsenden: As an example of the simplification mentioned above, this code could be reduced to just a simple `return new RemediationsCacheWriter(...);`

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheWriter.java
```diff
@@ -0,0 +1,337 @@
...
+        if (writer.initError != null) {
+            Exception cause = writer.initError;
+            writer.close();
```
rsenden: With the mechanism described above, there's no need to explicitly call `close()`; either `zipFs` was successfully initialized (in which case it will be automatically closed by outer try-with-resources`), or `zipFs` initialization failed (in which case there's nothing to close)

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheWriter.java
```diff
@@ -0,0 +1,337 @@
...
+        manifest.getEntries().add(toManifestEntry(order, entryPath, artifactId, releaseId, uploadDate, sha256));
+    }
+
+    /**
+     * Writes {@code manifest.json} into the open zip. Must be called before {@link #close()}
+     * for a successful cache. The archive is only published to {@code destination} after a
+     * successful {@link #close()}.
+     * Returns the completed manifest.
+     */
+    public RemediationsCacheManifest finish() {
```
rsenden: Why have this as a separate public method that needs to be explicitly called? Why not just have the `close()` method automatically perform all finalization operations (but through separate private helper methods, to keep `close()` method fairly short)

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheWriter.java
```diff
@@ -0,0 +1,337 @@
...
+                Files.move(workPath, destination, StandardCopyOption.REPLACE_EXISTING);
+            }
+        } catch (IOException e) {
+            deleteQuietly(workPath);
+            throw new FcliTechnicalException(
+                    "Failed to publish remediations cache zip to " + destination, e);
+        }
+    }
+
+    private void requireWritable() {
```
rsenden: Based on comments above, I think there's little use for this method

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheWriter.java
```diff
@@ -0,0 +1,337 @@
...
+            return;
+        }
+        try {
+            Files.deleteIfExists(path);
+        } catch (IOException e) {
+            logger.warn("Failed to delete incomplete remediations cache work file: {}", path, e);
+        }
+    }
+
+    public record FprSource(Path path, String artifactId, String releaseId, String uploadDate) {
```
rsenden: Having a single record that may represent two types of entities (SSC artifact id vs FoD release id) looks like a code smell; any way to improve?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/mixin/AviatorSSCApplyRemediationsSourceMixin.java
```diff
@@ -0,0 +1,102 @@
...
+import kong.unirest.UnirestInstance;
+import lombok.Getter;
+import picocli.CommandLine.ArgGroup;
+import picocli.CommandLine.Option;
+
+/**
+ * Source selection for apply-remediations: either online SSC selection or a local remediations cache zip.
+ */
+@Getter
+public class AviatorSSCApplyRemediationsSourceMixin {
```
rsenden, line 29: There seems to be quite some code duplication (like getter methods) between this class, `AviatorSSCRemediationsCacheDownloadSelectorMixin`, and possibly `AviatorSSCRemediationsSelectorArgGroups `; any way to improve? Given that these classes are now by themselves responsible for resolving artifact descriptors, do we still need these accessors? If so, can we use `@Getter` (where appropriate) instead?

rsenden (changes requested):

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheEntry.java
```diff
@@ -0,0 +1,50 @@
...
+    private String releaseId;
+    private String uploadDate;
+    private String path;
+    private String sha256;
+
+    /**
+     * Factory for writers (not Jackson). Avoids multi-arg same-type call sites on the wire model.
+     * Do not add {@code @Builder} on this {@code @Reflectable} class.
+     */
+    public static RemediationsCacheEntry of(
```
rsenden: Although it's an improvement to move this responsibility to the appropriate class:
- Given the number of arguments, most with the same type, there's still the risk that arguments are passed in the wrong order
- I don't like individual SSC and FoD properties being mixed in a single class; this makes it difficult to understand which properties are always required, and which properties are only applicable to either FoD or SSC

Before reading the below, please check the later comment about refactoring validation, as this is closely related and referenced in the proposals below.

A potentially better approach would be to have:
- Separate `RemediationsCacheFoDEntry` and `RemediationsCacheSSCEntry` classes (please check correct case for `FoD` and `SSC` in these class names to be consistent with other fcli classes), each containing only the relevant properties for either FoD or SSC
- Have both classes implement interface/abstract class that defines the shared properties like `order`, `path`, and `sha256`
- If SSC/FoD-specific properties may not be null, add a `validate` method to each entry class
- In `RemediationsCacheManifest`, declare two separate entry lists, i.e., `fodEntries` and `sscEntries`, or maybe keep single `List<? extends ...> entries`, deriving actual type from the `product` property

Alternatively, potentially easier from a caller perspective:
- Keep a single `RemediationsCacheEntry` class, but introduce two (inner) model classes that represent FoD/SSC-specific properties
- Introduce corresponding properties and accessors in this class, for example `sscData` and `fodData` (any better property names?)
- Validation would require exactly one of these properties to be set
- Validation of individual SSC/FoD-specific properties is done in the respective SSC/FoD-specific model classes
- Instead of the single `of` method, introduce separate factory methods for SSC/FoD, each taking the SSC/FoD-specific model class (not individual properties), just assigning the provided model object to the `sscData` or `fodData` field
- If useful, introduce factory methods on the SSC/FoD-specific model classes, like `of(artifactId, ...)` on the SSC model class, and `of(releaseId, ...)` on the FoD model class

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheReader.java
```diff
@@ -0,0 +1,230 @@
...
+        } catch (AbstractFcliException e) {
+            throw e;
+        } catch (IOException e) {
+            throw new FcliTechnicalException("Failed to read remediations cache manifest: " + cacheZip, e);
+        } catch (RuntimeException e) {
+            throw new FcliTechnicalException("Failed to parse remediations cache manifest: " + cacheZip, e);
+        }
+    }
+
+    private void validateManifest(RemediationsCacheManifest manifest) {
```
rsenden: I think validation would be better moved to the model classes, i.e., this class would just check that `manifest` is not `null`, then call `manifest.validate(cacheZip)`, which checks its own property values, then calls `getEntries().forEach(e->e.validate(cacheZip))`. This provides better encapsulation, reducing the risk of missing any validation checks, and based on the suggestion above about having different model classes for SSC/FoD, this would then also automatically allow SSC/FoD-specific properties to be validated.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/mixin/AviatorSSCRemediationsSelectorArgGroups.java
```diff
@@ -0,0 +1,139 @@
...
+        public String getAppVersionId(UnirestInstance unirest) {
+            if (StringUtils.isBlank(appVersionNameOrId)) {
+                return null;
+            }
+            SSCAppVersionDescriptor descriptor = SSCAppVersionHelper.getRequiredAppVersion(
+                    unirest, appVersionNameOrId, getDelimiter(), "id");
+            return descriptor.getVersionId();
+        }
+
+        public String getSelectionMode() {
```
rsenden: I haven't checked whether the values produced by this method are only used in fcli output (in which case current implementation may be fine), or referenced/compared in any other code (in which case it would be better to use a proper `enum`)

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/helper/AviatorSSCApplyRemediationsHelper.java
```diff
@@ -12,77 +12,165 @@
...
+                .fprsSkipped(applyResult.skipped())
+                .totalRemediation(aggregated.totalRemediations())
+                .appliedRemediation(aggregated.appliedRemediations())
+                .skippedRemediation(aggregated.skippedRemediations())
+                .modifiedFiles(aggregated.modifiedFiles())
+                .action(RemediationsApplyHelper.actionLabel(aggregated))
+                .build());
+    }
+
+    private static ObjectNode toOnlineJson(OnlineResultData data) {
```
rsenden: Several observations/questions on this method and related methods/classes:
- In general, we should attempt to have each fcli command output a consistent (sub)set of properties regardless of options/operating mode, for example to allow for proper, useful table output (even if that's not the default output format). Ideally, to ensure consistent output and allow for better code re-use, code/class structure should represent which properties are always being written (independent of operating mode), and what properties are only applicable to a certain operating mode, for example by moving shared properties to a common superclass that's inherited by both `CacheResultData` and `OnlineResultData`.
- Given that these `*ResultData` classes are (primarily) meant to represent output properties, why are we manually constructing an `ObjectNode` instead of using Jackson serialization?
- If Jackson serialization isn't possible for any reason, rendering JSON output should be the responsibility of the data class, i.e., having `*ResultData::toJson` methods, to provide better encapsulation and make it less likely (being closer to the actual property definitions) that any properties are missed.

rsenden (changes requested):

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/helper/AviatorSSCApplyRemediationsHelper.java
kireetivar: I structured the always-present fields in a shared `CommonApplyResult` and put online/cache-only fields on separate result types, with each type rendering via `toJson()`.

I didn’t use a shared superclass or Jackson `valueToTree` here because the output needs a few custom details: the action field must be named `__action__`, and values like file paths and string sets need to show up as simple JSON strings/arrays. `toJson()` keeps that next to the field definitions.

Happy to move to inheritance and/or Jackson if you prefer that style for fcli result DTOs.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsCacheEntry.java
kireetivar: I went with nested `sscData` / `fodData` and a single ordered entries list (each cache is one product; apply needs one order) rather than separate top-level SSC/FoD entry types and dual lists. Happy to switch if you still prefer that shape.

PR #1062, opened 2026-07-27
feat: Aviator Connection Diagnose command 

kireetivar:
## Summary

Adds `fcli aviator connection diagnose` to help users troubleshoot Fortify Aviator connectivity from endpoint configuration through credential validation.

The command runs ordered diagnostic stages for endpoint parsing, DNS, TCP, optional proxy CONNECT, TLS/HTTP2, gRPC reachability, and optional credential checks. It supports diagnosing a direct `--url`, `--url` with `--token`, a saved `--aviator-session`, or an `--admin-config`, and returns structured table/JSON output with stage status, guidance, and evidence.

## Changes

- Add the `fcli aviator connection diagnose` command under `fcli aviator connection`.
- Add Aviator diagnostic model/probe classes for endpoint, DNS, TCP, proxy, TLS, and gRPC checks.
- Add optional token/admin credential validation after transport diagnostics succeed.
- Add token text resolution helper so `--token` supports `file:`, `string:`, and `env:` sources consistently.
- Add focused unit coverage for diagnostic staging, source handling, token resolution, command behavior, and gRPC connection helper behavior.

## Trade-offs

Credential validation failures are reported as optional diagnostic failures so users still get complete transport diagnostics and only required transport failures drive a non-zero process exit.

The TCP stage intentionally opens a separate connection from the proxy/TLS tunnel stage so next-hop TCP failures remain distinguishable from proxy CONNECT and TLS handshake failures.

## Output

1. Happy path
image

2. Invalid URL:
image

3. DNS fail:
image

4. TCP Fail:
image

5. http page not grpc:
image

6. with `--aviator-session`:
image

7. with `--url` and `--token`:
image

8. With `--admin-config`:
image


fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/connection/cli/cmd/AviatorConnectionDiagnoseCommand.java
```diff
@@ -0,0 +1,94 @@
...
+    private final AviatorConnectionDiagnoseHelper diagnoseHelper = new AviatorConnectionDiagnoseHelper();
+    private DiagnoseRunResult runResult;
+
+    /**
+     * Overrides {@link AbstractOutputCommand#call()} so a required stage failure
+     * yields exit code 1 after the diagnostic table is written. Soft status cannot
+     * use exceptions without dropping structured stage output.
+     */
+    @Override
+    public Integer call() {
```
rsenden: The `AbstractOutputCommand::call` method is not meant to be overridden; not having this declared as `final` is a code bug. I've made some changes in `dev/v3.x` branch to allow `AbstractOutputCommand` implementations to generate a custom exit code.

After merging these changes into the Aviator branch, the intended behavior can be implemented using something like the following:

```
@Override
protected IObjectNodeProducer getObjectNodeProducer() {
    if (timeoutSeconds <= 0) {
        throw new FcliSimpleException("--timeout must be greater than 0");
    }
    var runResult = new AviatorConnectionDiagnoseHelper().diagnose(resolveSource(), timeoutSeconds);
    return simpleObjectNodeProducerBuilder(ObjectNodeProducerApplyFrom.SPEC)
            .source(runResult.json())
            .exitCode(runResult.requiredFailure() ? 1 : 0)
            .build();
}
```

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/connection/cli/cmd/AviatorConnectionDiagnoseCommand.java
```diff
@@ -0,0 +1,94 @@
...
+public class AviatorConnectionDiagnoseCommand extends AbstractOutputCommand implements IJsonNodeSupplier {
+    @Getter @Mixin private OutputHelperMixins.TableNoQuery outputHelper;
+
+    @ArgGroup(exclusive = true, multiplicity = "1", headingKey = "aviator.connection.diagnose.source.arggroup")
+    private AviatorConnectionDiagnoseSourceArgGroup sourceArgGroup;
+
+    @Option(names = "--timeout", defaultValue = "30", paramLabel = "<seconds>")
+    private int timeoutSeconds;
+
+    private final AviatorConnectionDiagnoseHelper diagnoseHelper = new AviatorConnectionDiagnoseHelper();
```
rsenden: As per fcli conventions, instance fields in command classes are only allowed for picocli options/mixins/..., not for storing data/utilities for use across method implementations, like `diagnoseHelper` and `runResult`. See comment below on how this can be resolved.

rsenden (changes requested):
Apart from individual code comments, I'm having doubts about using 3rd-party domains in tests, like `aviator.example.com`, especially if these tests actually attempt to connect to these 3rd-party domains. This can expose private data like IP-addresses of developer workstations to those 3rd-party domains, and they may not like us using their domains to run tests. Can we change this to use non-existing or OpenText-owned domains?

PR #1066, opened 2026-07-31
feat: Add Aviator source encoding handling 

kireetivar:
## Summary

- Add configurable Aviator source file encoding handling through `--source-encodings`, defaulting to `FPR,UTF-8,ISO-8859-1`.
- Use FPR source encoding metadata when decoding source files, with fallback encoding candidates for audit and remediation flows.
- Propagate source encoding options through Aviator SSC audit, SSC/FoD apply-remediations, FPR processing, source enrichment, and gRPC calls.
- Improve Aviator audit results by reporting skipped issue counts, skipped reasons, and remediation-generation skip metrics.

```text
feat: `fcli aviator ssc audit`: Add `--source-encodings` option for source decoding and audit skip reporting
feat: `fcli aviator ssc apply-remediations`: Add `--source-encodings` option for source file decoding and encoding
feat: `fcli fod aviator apply-remediations`: Add `--source-encodings` option for source file decoding and encoding
```


rsenden (changes requested):
I haven't reviewed everything in detail, focusing on overall structure first. Instead of having custom option parsing functionality, it would be much better to use existing picocli functionality; something like the following:
- Create `ISourceDecoder` interface, with single `decode` method matching signature of current `SourceEncodingOptions::decode`
- Create implementations of this interface:
    - One implementation in which the `decode` method retrieves encoding from FVDL
    - One that takes a fixed `Charset` as constructor arg (together with a String-based constructor that calls `Charset.forName`)
    - Optionally a `CompositeSourceDecoder` that takes a `List<ISourceDecoder>` as constructor argument
- Create a corresponding picocli `ITypeConverter` that based on input (`FPR` or explicit charset) returns an instance of one of the classes above, throwing a proper picocli exception like `TypeConversionException` to have picocli render a proper error message in case of invalid charset
- Create picocli mixin/arggroup class (ideally in `fcli-aviator-common` for re-use across FoD/SSC, but not sure whether we currently have any picocli-related classes in this module) that defines `--source-encodings` as `List<ISourceDecoder>` (or directly as `CompositeSourceDecoder`, but not sure whether that would work with `split`), with proper `split`, `converter`, and `defaultValue` attributes
- Pass this list of candidate decoders to the processors (either as-is, or wrapped in `CompositeSourceDecoder`, but for flexibility, processors would rely on `ISourceDecoder` interface)
- For encoding, have a separate `SourceEncoder` class with static `encode` method, as this operates on specified `Charset` and doesn't (directly) rely on any of the above

rsenden (approved):

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/audit/model/AuditResponse.java
```diff
@@ -26,13 +28,76 @@
...
+        UNKNOWN(null, "Unknown audit failure", null),
+        OTHER(null, null, null);
+
+        private static final String CLIENT_SIDE_ERROR_PREFIX = "Client-side pre-processing error: ";
+
+        private final String messageFormat;
+        private final String displayMessage;
+        private final Pattern messagePattern;
+
+        AuditSkipReason(String messageFormat, String displayMessage, Pattern messagePattern) {
```
rsenden: Please consider Lombok `@RequiredArgsConstructor`

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/audit/model/AuditResponse.java
```diff
@@ -43,6 +108,11 @@ public class AuditResponse {
     private String userPrompt;
     private String systemPrompt;
 
+    @JsonIgnore
+    public AuditSkipReason getAuditSkipReason() {
+        return auditSkipReason == null ? AuditSkipReason.from(status, statusMessage) : auditSkipReason;
+    }
+
     public AuditResponse(AuditResult auditResult, int inputToken, int outputToken, String status,
```
rsenden: This code was already present before this PR, but this constructor takes way too many arguments, thus easy to get the order wrong. Please consider (Lombok) builder pattern instead (for this or future PR).

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/audit/model/AuditResponse.java
```diff
@@ -26,13 +28,76 @@
...
+        public String format(Object... args) {
+            return String.format(messageFormat, args);
+        }
+
+        public static AuditSkipReason from(String status, String statusMessage) {
+            String message = effectiveMessage(status, statusMessage);
+            if (message.isBlank()) {
+                return UNKNOWN;
+            }
+            for (AuditSkipReason reason : values()) {
```
rsenden: This is still doing string-based matching; matching has just been updated from string comparison to pattern matching. It would be better to simply store the `AuditSkipReason` enum value in `AuditResponse`, such that you can directly access the appropriate enum entry and associated display message. That might also remove the need for having `UNKNOWN` and/or `OTHER` enum values.

rsenden (changes requested):

Issue #1058, opened 2026-07-16
FCLI version 3.23.3 error when triggering DAST Automated


mramos3-OT:
### Current Behavior

Error: 
```
FcliSimpleException: Unable to start Dynamic scan after 30 attempts. Please check the UI and try again
at com.fortify.cli.fod._common.scan.helper.dast.FoDScanDastAutomatedHelper.handleInProgressScan(FoDScanDastAutomatedHelper.java:107)
```

### Expected Behavior

DAST Automated via `fcli fod dast-scan start` should trigger the scan

### Steps To Reproduce

Able to replicate through PS CLI with version 3.23.2 with the below command (see attached graphic):

- `fcli fod dast-scan start --release [releaseID]`

### Environment

```markdown
2026-07-03T07:07:35.7747613Z ##[section]Starting: Trigger Scan
2026-07-03T07:07:35.7756584Z ==============================================================================
2026-07-03T07:07:35.7756724Z Task         : PowerShell
2026-07-03T07:07:35.7756806Z Description  : Run a PowerShell script on Linux, macOS, or Windows
2026-07-03T07:07:35.7756911Z Version      : 2.276.0
2026-07-03T07:07:35.7756999Z Author       : Microsoft Corporation
2026-07-03T07:07:35.7757078Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/powershell
2026-07-03T07:07:35.7757195Z ==============================================================================
2026-07-03T07:07:37.0645751Z Generating script.
2026-07-03T07:07:37.1070798Z ========================== Starting Command Output ===========================
2026-07-03T07:07:37.1347742Z ##[command]"C:\Program Files\PowerShell\7\pwsh.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'C:\a\_temp\207f8c8e-5b9b-4277-bdf1-60e2c0104063.ps1'"
2026-07-03T07:07:38.1092831Z Logging into Fortify on Demand...
2026-07-03T07:07:39.2075749Z  Name     Type    Url                          Created                  Expires                  Expired  Action  
2026-07-03T07:07:39.2077128Z  default  FoD     https://api.ams.fortify.com  2026-07-03 07:07:39 UTC  2026-07-03 13:07:34 UTC  No       CREATED 
2026-07-03T07:07:39.2077587Z 
2026-07-03T07:07:39.2077821Z 
[...]

PR #1068, opened 2026-08-04
fix: start DAST scan on release with no prior scans

jmadhur87:
fcli fod dast-scan start threw FcliSimpleException: Unable to start Dynamic scan after 30 attempts on releases that had no previous scans.

Changed the empty-items guard from continue to return null. An empty scan list means no active scan exists for the release, so the method now immediately signals that a new scan can proceed.

fix: start DAST scan on release with no prior scans(#1058 )


rsenden:
Superseded by https://github.com/fortify/fcli/pull/1070

PR #1071, opened 2026-08-11
fix: reject corrupted Windows remediation paths

kireetivar:
Windows command-line decoding can corrupt non-ASCII path characters before fcli receives them. For example, characters in paths such as `Fortify 测试` or `Android ऐप` may arrive as `?` or `U+FFFD`.

This can cause remediation cache commands to use an invalid path or fail with an unclear error.

## Changes

- Added shared Windows path validation for `?` and the Unicode replacement character (`U+FFFD`).
- Added Picocli converters for validated `Path` and `File` values.
- Applied validation to SSC and FoD remediation commands:
  - `--from-cache`
  - `-f/--file`
  - `--source-dir`
- Added regression tests for corrupted cache paths, output files, and source directories.
- Removed the need for native JNA-based command-line recovery.
- Kept validation Windows-only so Linux and macOS path behavior remains unchanged.


rsenden:
As discussed, we'll defer corrupted CLI arguments handling for now.

PR #1073, opened 2026-08-13
feat: `fcli fod session login`: Add support for TOTP and multi-factor authentication (MFA) codes in the FoD session login command

SangameshV:
### Summary:
Adds support for TOTP and Multi-Factor Authentication (MFA) security codes during FoD session authentication. Users can now provide a security code as part of the FoD session login command when their FoD tenant requires additional authentication.

### New CLI Options:
- `--code (-c)`: Optional parameter to provide a TOTP or MFA security code. Can be used interactively or as a command-line argument.
- `--totp`: Optional flag indicating that the provided code is a TOTP generated by an authenticator application. If omitted, the code is treated as a standard MFA code (for example, a code received via email).

### Tested the following scenarios:
1. When the FoD tenant requires MFA or TOTP authentication and no security code is supplied, the login command now returns a clear and actionable error message explaining how to provide the required code.
2. Users can authenticate successfully by providing a valid MFA code received through supported channels.
3. If the provided security code is invalid, expired, or otherwise rejected by FoD, the login command returns an improved error message with guidance on retrying authentication.

### Benefits:
- Enables authentication against FoD tenants that enforce MFA or TOTP verification.
- Improves user experience with clearer authentication guidance.
- Provides actionable error messages for missing, invalid, or expired security codes.
- Maintains backward compatibility for tenants that do not require additional authentication.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionLoginCommand.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionLoginCommand.java
@@ -25,8 +25,6 @@ import com.fortify.cli.fod._common.session.helper.FoDSessionDescriptor;
 import com.fortify.cli.fod._common.session.helper.FoDSessionHelper;
 import com.fortify.cli.fod._common.session.helper.oauth.FoDOAuthHelper;
 import com.fortify.cli.fod._common.session.helper.oauth.FoDTokenCreateResponse;
-import com.fortify.cli.fod._common.session.helper.oauth.impl.BasicFoDUserAuthCode;
-import com.fortify.cli.fod._common.session.helper.oauth.impl.BasicFoDUserCredentials;
 
 import lombok.Getter;
 import picocli.CommandLine.Command;
@@ -38,6 +36,25 @@ public class FoDSessionLoginCommand extends AbstractSessionLoginCommand<FoDSessi
     @Getter private FoDSessionHelper sessionHelper = FoDSessionHelper.instance();
     @Mixin private FoDSessionLoginOptions loginOptions;
     @Mixin private FoDUnirestInstanceSupplierMixin unirestInstanceSupplierMixin;
+
+    private static final String MFA_GUIDANCE = "If MFA is required, provide the security code:\n"
+            + "  --code <code>  (or -c <code>) to provide the security code\n"
+            + "  --totp          to indicate the code is from a TOTP authenticator app";
+
+    private static final String ERROR_WITH_CODE = "Authentication failed. Possible causes:\n"
+            + "  - Incorrect username or password\n"
+            + "  - MFA/TOTP code incorrect, expired, or wrong type (TOTP vs MFA)\n"
+            + "Please verify your credentials and MFA/TOTP code if applicable:\n"
+            + MFA_GUIDANCE;
+
+    private static final String ERROR_WITHOUT_CODE = "Authentication failed. Possible causes:\n"
+            + "  - Incorrect username or password\n"
+            + "  - FoD tenant requires MFA/TOTP authentication\n\n"
+            + MFA_GUIDANCE;
+
+    private static final String ERROR_CLIENT_CREDENTIALS = "Authentication failed. Possible causes:\n"
+            + "  - Incorrect client ID or client secret\n"
+            + "  - Client credentials have expired";
 
     @Override
     public ISessionNameSupplier getSessionNameSupplier() {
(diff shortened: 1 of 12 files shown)


fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionLoginCommand.java
```diff
@@ -50,12 +54,43 @@ protected void logoutBeforeNewLogin(String sessionName, FoDSessionDescriptor ses
...
         IUrlConfig urlConfig = loginOptions.getUrlConfigOptions();
-        if ( loginOptions.hasClientCredentials() ) {
+        if (loginOptions.hasClientCredentials()) {
             FoDTokenCreateResponse createTokenResponse = FoDOAuthHelper.createToken(urlConfig, loginOptions.getClientCredentialOptions(), loginOptions.getAuthOptions().getScopes());
             sessionDescriptor = new FoDSessionDescriptor(urlConfig, createTokenResponse);
-        } else if ( loginOptions.hasUserCredentials() ) {
-            FoDTokenCreateResponse createTokenResponse = FoDOAuthHelper.createToken(urlConfig, loginOptions.getUserCredentials(), loginOptions.getAuthOptions().getScopes());
-            sessionDescriptor = new FoDSessionDescriptor(urlConfig, createTokenResponse);
+        } else if (loginOptions.hasUserCredentials()) {
+            var credBuilder = BasicFoDUserCredentials.builder()
```
rsenden: Why did you move this code for building credential objects into this class? The original code was much more compact with better separation of concerns. With the new MFA support, I'd expect a single line like this to create the token:

```
FoDTokenCreateResponse createTokenResponse = FoDOAuthHelper.createToken(urlConfig, loginOptions.getUserCredentials(), loginOptions.getMfaCode(), loginOptions.getAuthOptions().getScopes());
```

`FoDOAuthHelper` should be null-safe for MFA code, i.e., it should handle both full object is `null`, or non-null object with `securityCode` being `null`. For `loginOptions::getMfaCode`, probably best to return `null` object if `securityCode` option value is `null`.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionLoginCommand.java
```diff
@@ -50,12 +54,43 @@ protected void logoutBeforeNewLogin(String sessionName, FoDSessionDescriptor ses
...
+                    .password(loginOptions.getUserCredentialOptions().getPassword());
+
+            var authCodeBuilder = BasicFoDUserAuthCode.builder()
+                    .securityCode(loginOptions.getUserCredentialOptions().getSecurityCode())
+                    .isTotp(loginOptions.getUserCredentialOptions().isTotp());
+            try {
+                FoDTokenCreateResponse createTokenResponse = FoDOAuthHelper.createToken(urlConfig, credBuilder.build(),
+                        authCodeBuilder.build(), loginOptions.getAuthOptions().getScopes());
+                sessionDescriptor = new FoDSessionDescriptor(urlConfig, createTokenResponse);
+            } catch (UnexpectedHttpResponseException e) {
```
rsenden, line 87: - Current exception handling block is too long, taking away attention from actual logic above. Closely related, it doesn't make sense to construct three local strings with static contents; better to move these to actual constants, which would automatically make exception handling code much shorter
- Why do we have dedicated exception handling when authenticating with user credentials, but not when authenticating with client credentials? Shouldn't we throw a similar `FcliSimpleException` with appropriate guidance (like client credentials incorrect or expired)?

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/mixin/FoDSessionLoginOptions.java
```diff
@@ -89,55 +95,47 @@ public final boolean hasUserCredentials() {
...
         @Override
         protected int getDefaultSocketTimeoutInMillis() {
             return 600000;
         }
     }
-
-    /**
-     * Basic immutable FoD user credentials with builder pattern.
-     */
-    public static final class BasicFoDUserCredentials implements IFoDUserCredentials {
```
rsenden, line 120: I think (but please double-check) that `BasicFoDUserCredentials` and the new `BasicFoDUserAuthCode` are currently only referenced within `FoDSessionLoginOptions`, so from that perspective, it makes more sense to keep these as inner classes instead of separate top-level types. However, the HTTP MCP server for example defines its own `HttpMcpFoDUserCredentials` class with similar implementation as `BasicFoDUserCredentials`. As such, it may makes sense to keep the new top-level `BasicFoD*` classes, and change the MCP HTTP server (and any other code that creates `IFoDUserCredentials`) to utilize these top-level `BasicFoD*` classes instead of providing their own implementations of `IFoDUserCredentials`.

Can you please check:
- Which classes currently implement `IFoDUserCredentials` and similar interfaces (like SSC variant)?
- Whether these can be easily replaced with a single, top-level `Basic*` implementation

To summarize:
- If existing implementations of `IFoDUserCredentials` can be easily merged, keep the top-level `Basic*` classes and refactor all code to use this single implementation
- If there are significant differences between implementations and `Basic*` classes are only instantiated in `FoDSessionLoginOptions`, move those `Basic*` classes back to `FoDSessionLoginOptions` as inner classes

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/helper/oauth/FoDOAuthHelper.java
```diff
@@ -28,8 +28,16 @@
 // TODO Consider moving all classes in this package to a more appropriate package,
 //      for example as a sub-package of the 'rest' package.
 public class FoDOAuthHelper {
+    
     public static final FoDTokenCreateResponse createToken(IUrlConfig urlConfig, IFoDUserCredentials uc, String... scopes) {
-        Map<String,Object> formData = generateTokenRequest(uc, scopes);
+        Map<String,Object> formData = generateTokenRequest(uc, IFoDUserAuthCode.NONE, scopes);
+    try ( var unirest = UnirestHelper.createUnirestInstance() ) {
```
rsenden: - Indentation seems wrong (looks like this was already wrong in original code, but please fix)

rsenden (changes requested):

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionLoginCommand.java
SangameshV: I have had class-level constants created for the messages and have used them. 

Note: Across the entire codebase, every other command throws `FcliSimpleException` with the message string inline at the throw site, either as a string literal directly or via `String.format(...)`. No other command pre-defines multi-line message constants like `MFA_GUIDANCE`, `ERROR_WITH_CODE`, and `ERROR_WITHOUT_CODE`.

SangameshV: Added dedicated exception handling for client credentials as well.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/helper/oauth/FoDOAuthHelper.java
SangameshV: Done

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/mixin/FoDSessionLoginOptions.java
SangameshV: The Basic* classes are kept as top-level classes. please review the code now.

PR #1076, opened 2026-08-20
feat: add --preview flag to preview aviator remediations before applying

dhanwanthp:
This MR adds a new `--preview` flag to the fcli aviator ssc apply-remediations and fcli fod aviator apply-remediations commands, enabling users to preview what changes would be applied to their source code without actually modifying files.

## Features:
- Added --preview option to both SSC and FoD apply-remediations commands
- When enabled, the tool performs full validation and processing without modifying source files
- Returns detailed JSON output containing:
    - All proposed code changes per issue ID
    - File paths and encodings
    - Line numbers and code snippets (before/after)
    - Context metadata for fuzzy matching
    - Skip reasons for any failed remediations
    
## Usage: 
```
# Preview changes without applying them
fcli aviator ssc apply-remediations --appversion myapp --preview

# Preview specific issues only
fcli aviator ssc apply-remediations --appversion myapp --preview --issue-ids ISSUE-123,ISSUE-456

# Apply changes (existing behavior, no --preview flag)
fcli aviator ssc apply-remediations --appversion myapp
```

Output format:-
```
{
  "previewMode": true,
  "totalRemediation": 10,
  "appliedRemediation": 8,
  "previewDetails": [
    {
      "issueId": "ISSUE-123",
      "status": "available",
      "files": {
        "src/Example.java": {
          "filename": "src/Example.java",
          "path": "src/Example.java",
          "encoding": "UTF-8",
          "changes": [
            {
              "changeIndex": 1,
              "lineFrom": 42,
              "lineTo": 44,
              "originalCode": "...",
              "newCode": "...",
              "context": { "linesBefore": 2, "linesAfter": 2, "content": "..." },
              "fuzzyMatched": false
            }
          ]
        }
      }
    }
  ]
}
```


fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/remediations_cache/RemediationsApplyHelper.java
```diff
@@ -52,13 +52,27 @@ public static ApplyResult apply(
...
             Set<String> issueIdFilter,
             Logger skipLog) {
+        return apply(source, sourceCodeDirectory, logger, issueIdFilter, skipLog, false);
+    }
+
+    /**
+     * Applies or previews remediations for each source entry until done or the issue-id filter is exhausted.
+     * Caller owns {@code source} lifecycle (try-with-resources).
+     */
+    public static ApplyResult apply(
```
rsenden, line 50: As the number of parameters keep growing, maybe better to combine in a single object that implements builder pattern (through Lombok `@Builder`), i.e.,

```
  apply(ApplyArgs.builder()
              .source(source)
              .sourceCodeDirectory(sourceCodeDirectory)
              ...
              .build();
```

Same could potentially be applied to other methods in Aviator code to reduce number of parameters.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/processor/preview/FileChange.java
```diff
@@ -0,0 +1,52 @@
...
+ * @param lineFrom Starting line number (1-based, inclusive)
+ * @param lineTo Ending line number (1-based, inclusive)
+ * @param originalCode The code being replaced
+ * @param newCode The replacement code
+ * @param context Context lines surrounding the change (for validation)
+ * @param fuzzyMatched True if file hash didn't match and fuzzy context search was used
+ */
+@Reflectable
+@JsonPropertyOrder({"changeIndex", "lineFrom", "lineTo", "originalCode", "newCode", "context", "fuzzyMatched"})
+public record FileChange(
```
rsenden, line 36: Similar comment as before; maybe better to add Lombok `@Builder` instead of relying on constructor parameter order.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -100,7 +102,7 @@ private JsonNode processOnline(
         try (SSCOnlineRemediationsFprSource source = new SSCOnlineRemediationsFprSource(
                 unirest, logger, progressWriter, resolved.artifacts())) {
             ApplyResult applyResult = RemediationsApplyHelper.apply(
-                    source, sourceCodeDirectory, logger, issueIdFilter, LOG);
+                    source, sourceCodeDirectory, logger, issueIdFilter, LOG, previewMode);
```
rsenden: Related to the `apply` comment elsewhere, what about:
- Move the various options that define method parameter values into a separate Picocli ArgGroup that can be shared between SSC and FoD commands
- Have that ArgGroup implement some interface with getters for each of the option values (through Lombok `@Getter` annotations on the options/class)
- Have the `RemediationsApplyHelper::apply` method take an instance of that interface

This way, the FoD & SSC `apply-remediations` commands can simply pass the ArgGroup to the `apply` method (together with the loggers, which remain separate method parameters), and due to interface abstraction, `RemediationsApplyHelper` doesn't have a compile-time dependency on CLI-specific code (like the ArgGroup).

rsenden: Why are you passing `LOG` to a different class? Usually, each class has its own Slf4j logger, which also makes it much easier to identify which class generated a particular log message.

rsenden: I think there's some overlap between FoD and SSC implementations of the `apply-remediations` command; please check whether it makes sense to introduce a common `AbstractAviatorApplyRemediationsCommand` base class.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,74 @@
...
+        implements IJsonNodeSupplier, IRecordTransformer, IActionCommandResultSupplier {
+
+    @Getter @Mixin private OutputHelperMixins.DetailsNoQuery outputHelper;
+    @Mixin private ProgressWriterFactoryMixin progressWriterFactoryMixin;
+    @Mixin protected ApplyRemediationsOptionsMixin applyOptions;
+
+    @Override
+    public final JsonNode getJsonNode() {
+        validateSourceSelector();
+        AviatorApplyRemediationsCliSupport.requireSourceDir(applyOptions.getSourceCodeDirectory());
```
rsenden: Exceptions thrown by `AviatorApplyRemediationsCliSupport` reference explicit option names, whereas that class doesn't 'know' from which command class it's being invoked, and what options are provided by that command. The `requiresSourceDir` method can be easily moved to this abstract base class, as this base class declares the option for specifying source code directory. For `requireIssueIdsCacheOnly`, it's more difficult as this abstract base class doesn't declare the `--from-cache` option. I guess `--from-cache` is currently declared through picocli arg group as being exclusive to product-specific options, hence we can't move `--from-cache` to `ApplyRemediationsOptionsMixin` without loosing the picocli exclusivity. Given that this option is shared between SSC & FoD, maybe it's still worth declaring this option on this abstract base class, and manually check for exclusivity. 

Effectively, I think it would be good to get rid of the `AviatorApplyRemediationsCliSupport` class, moving functionality to this abstract base class.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,74 @@
...
+                    : processOnline(logger, progressWriter, issueIdFilter);
+        }
+    }
+
+    /** Override to validate source selector state before options are checked; no-op by default. */
+    protected void validateSourceSelector() {}
+
+    protected abstract boolean isFromCacheSelected();
+
+    protected abstract JsonNode processFromCache(AviatorLoggerImpl logger, Set<String> issueIdFilter);
```
rsenden: I haven't checked the details, but wouldn't applying from cache be very similar between SSC & FoD? Or, possibly even better, can't we get rid of separate methods for cache vs online, and instead have something like an `IFpr[s]Supplier` interface that, based on given options, provides FPR files from either cache, SSC, or FoD, with actual FPR processing logic shared between all three use cases?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/util/AviatorRemediationMetricsHelper.java
```diff
@@ -40,7 +40,18 @@ private AviatorRemediationMetricsHelper() {}
     public static RemediationMetric aggregateMetrics(Set<String> requestedIssueIds, Collection<RemediationMetric> metrics) {
```
rsenden, line 41: This seems a fairly long method; can we improve this through self-describing sub-methods, and/or through utility (builder) methods on `RemediationMetric`?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/processor/preview/ChangeDetail.java
```diff
@@ -0,0 +1,56 @@
...
+ * @param lineTo Ending line number (1-based)
+ * @param originalCode Code being replaced
+ * @param newCode Replacement code
+ * @param contextLinesBefore Number of context lines before the change
+ * @param contextLinesAfter Number of context lines after the change
+ * @param contextContent Full context text from remediations.xml
+ * @param fuzzyMatched Whether fuzzy matching was used
+ */
+@Reflectable
+public record ChangeDetail(
```
rsenden: Given the many record constructor arguments, can we use Lombok `@Builder` to improve code readability and avoiding constructor arguments potentially being passed in wrong order?

rsenden (changes requested):

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+    protected abstract IRemediationsFprSource openFprSource(AviatorLoggerImpl logger, IProgressWriter progressWriter);
+
+    protected abstract JsonNode buildResultNode(IRemediationsFprSource source, ApplyResult result, Set<String> issueIdFilter);
+
+    @Override
+    public final boolean isSingular() { return true; }
+
+    @Override
+    public JsonNode transformRecord(JsonNode record) { return record; }
```
rsenden: As this is a no-op, better to remove this method and the corresponding `IRecordTransformer` interface from the class definition.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private void requireSourceDir() {
```
rsenden: Given the default value for `--source-dir`, I don't think the `null` check is necessary? Or better yet, just use `StringUtils::isBlank`, which handles both `null` and blank values. Shouldn't we also validate that the path exists/is accessible? Given that exception explicitly mentions option name, it would be better to move this validation to the mixin that declares this option. This could be done either by an explicit validation method on the mixin, or directly in `getSourceCodeDirectory` (i.e., checks are done when value is retrieved, although pre-flight validation may be better).

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,85 @@
...
+        FcliSimpleException.throwIf(applyOptions.getSourceCodeDirectory() == null || applyOptions.getSourceCodeDirectory().isBlank(),
+                "--source-dir must specify a valid directory path");
+    }
+
+    private void requireIssueIdsCacheOnly() {
```
rsenden: Similar to the above; as exception references option names, better to move this validation to the mixin.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/aviator/cmd/FoDAviatorApplyRemediationsCommand.java
```diff
@@ -40,76 +30,37 @@
...
-
-    @Override
-    public String getActionCommandResult() {
-        return "Remediation-Applied";
-    }
-
-    @Override
-    public JsonNode transformRecord(JsonNode record) {
-        return record;
+        return AviatorFoDApplyRemediationsHelper.buildOnlineResultNode(resolvedRelease, result);
```
rsenden: Why is this done through `IRecordTransformer` instead of directly in the `buildResultNode` method?

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public final JsonNode getJsonNode() {
+        validateSourceSelector();
+        requireSourceDir();
+        requireIssueIdsCacheOnly();
+        Set<String> issueIdFilter = AviatorIssueIdFilterUtils.normalizeIssueIds(applyOptions.getIssueIds());
+        try (IProgressWriter progressWriter = progressWriterFactoryMixin.create()) {
+            AviatorLoggerImpl logger = new AviatorLoggerImpl(progressWriter);
+            try (IRemediationsFprSource source = openFprSource(logger, progressWriter)) {
```
rsenden: `source` can be easily mistaken for 'source file/dir'; can we consistently rename this variable to `fprSource` (also in for example `RemediationsApplyHelper` if applicable)

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/_common/output/cli/cmd/AbstractAviatorApplyRemediationsCommand.java
```diff
@@ -0,0 +1,85 @@
...
+import com.fortify.cli.common.output.cli.cmd.IJsonNodeSupplier;
+import com.fortify.cli.common.output.cli.mixin.OutputHelperMixins;
+import com.fortify.cli.common.output.transform.IRecordTransformer;
+import com.fortify.cli.common.progress.cli.mixin.ProgressWriterFactoryMixin;
+import com.fortify.cli.common.progress.helper.IProgressWriter;
+
+import lombok.Getter;
+import picocli.CommandLine.Mixin;
+
+public abstract class AbstractAviatorApplyRemediationsCommand extends AbstractOutputCommand
```
rsenden, line 37: Although much better than before, I think definitions and responsibilities between abstract base class, concrete sub-classes, and options/mixins can be further improved.

Maybe something like the following?
- Rename `ApplyRemediationsOptionsMixin` to `AbstractApplyRemediationsOptionsMixin`
- Add `validate` method to this abstract mixin and its interface, calling individual validation methods for validating source selection, source dir, `--issue-ids`/`--from-cache` interdependency, ...
- Have both SSC & FoD remediation mixins extend the abstract mixin
- Declare remediation mixin in concrete FoD/SSC command class, accessible to abstract command class through getter

So, basically, generic & command-specific options are provided through single mixin on FoD/SSC command classes, and those mixins provide the necessary validation logic. This allows the abstract base class to focus on process, instead of handling both process and data (validation).

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -13,112 +13,63 @@
...
-    @Mixin private ProgressWriterFactoryMixin progressWriterFactoryMixin;
+public class AviatorSSCApplyRemediationsCommand extends AbstractAviatorApplyRemediationsCommand {
     @Mixin private AviatorSSCApplyRemediationsSourceMixin sourceSelector;
     @Mixin private SSCUnirestInstanceSupplierMixin unirestInstanceSupplier;
-
-    @Option(names = {"--source-dir"})
-    private String sourceCodeDirectory = System.getProperty("user.dir");
-    @Option(names = {"--issue-ids"}, split = ",")
-    private List<String> issueIds;
+    private ResolvedOnlineArtifacts resolvedOnline;
```
rsenden: As per fcli convention, command classes ideally shouldn't store state in instance fields

rsenden (changes requested):

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -13,112 +13,60 @@
...
-                issueIds, sourceSelector.isFromCacheSelected());
-
-        try (IProgressWriter progressWriter = progressWriterFactoryMixin.create()) {
-            AviatorLoggerImpl logger = new AviatorLoggerImpl(progressWriter);
-            if (sourceSelector.isFromCacheSelected()) {
-                return processFromCache(logger, issueIdFilter);
-            }
-            return processOnline(logger, progressWriter, issueIdFilter);
-        }
+    protected AbstractApplyRemediationsOptionsMixin getApplyOptions() {
```
rsenden: I think just adding `@Getter` on the field will work as well?

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/mixin/SscApplyRemediationsOptionsMixin.java
```diff
@@ -0,0 +1,38 @@
...
+
+import lombok.Getter;
+import picocli.CommandLine.Mixin;
+
+/**
+ * SSC-specific apply-remediations options mixin. Composes SSC source selection with shared options
+ * and provides SSC-specific validation logic.
+ */
+@Getter
+public final class SscApplyRemediationsOptionsMixin extends AbstractApplyRemediationsOptionsMixin {
```
rsenden: Why did you introduce a new mixin class (which unlike other mixins doesn't have the `Aviator` prefix)? Wouldn't the following be better:
- Rename `AviatorSSCApplyRemediationsSourceMixin` to `AviatorSSCApplyRemediationsOptionsMixin`
- Have `AviatorSSCApplyRemediationsOptionsMixin` extend `AbstractApplyRemediationsOptionsMixin` and implement `validateSourceSelection` and `isCacheMode` methods
- Remove `SscApplyRemediationsOptionsMixin `

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/aviator/cli/mixin/FoDApplyRemediationsOptionsMixin.java
```diff
@@ -0,0 +1,46 @@
...
+
+import lombok.Getter;
+import picocli.CommandLine.Mixin;
+
+/**
+ * FoD-specific apply-remediations options mixin. Composes FoD source selection with shared options
+ * and provides FoD-specific validation logic. Implements delimiter injection for source selector.
+ */
+@Getter
+public final class FoDApplyRemediationsOptionsMixin extends AbstractApplyRemediationsOptionsMixin
```
rsenden: Same comment as for SSC

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/aviator/cmd/FoDAviatorApplyRemediationsCommand.java
```diff
@@ -12,104 +12,58 @@
...
-                issueIds, sourceSelector.isFromCacheSelected());
-
-        try (IProgressWriter progressWriter = progressWriterFactoryMixin.create()) {
-            AviatorLoggerImpl logger = new AviatorLoggerImpl(progressWriter);
-            if (sourceSelector.isFromCacheSelected()) {
-                return processFromCache(logger, issueIdFilter);
-            }
-            return processOnline(logger, issueIdFilter);
-        }
+    protected AbstractApplyRemediationsOptionsMixin getApplyOptions() {
```
rsenden: Same comment as for SSC (use `@Getter`)

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -13,112 +13,60 @@
...
-            ApplyResult applyResult = RemediationsApplyHelper.apply(
-                    source, sourceCodeDirectory, logger, issueIdFilter, LOG);
-            return AviatorSSCApplyRemediationsHelper.buildCacheResultNode(
-                    sourceSelector.getFromCache(),
-                    applyResult,
-                    issueIdFilter,
-                    source.reader().getManifest().getSelection());
+    @Override
+    protected IRemediationsFprSource openFprSource(AviatorLoggerImpl logger, IProgressWriter progressWriter) {
+        if (applyOptions.getSourceSelector().isFromCacheSelected()) {
```
rsenden: I think code becomes more readable with an explicit `else` block, making it more clear that these are two exclusive paths. Better yet, introduce two new methods, and change this method to:

```
@Override
protected IRemediationsFprSource openFprSource(AviatorLoggerImpl logger, IProgressWriter progressWriter) {
    return applyOptions.isFromCacheSelected() 
        ? openCacheFprSource(...)
        : openOnlineFprSource(...);
}
```

Note that the same concern applies to `buildResultNode` in this class (i.e., introduce `buildCacheResultNode` and `buildOnlineResultNode`), and do the same changes in the FoD command.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/aviator/cmd/FoDAviatorApplyRemediationsCommand.java
```diff
@@ -12,104 +12,58 @@
...
-    private JsonNode processOnline(AviatorLoggerImpl logger, Set<String> issueIdFilter) {
-        UnirestInstance unirest = unirestInstanceSupplier.getUnirestInstance();
-        FoDReleaseDescriptor release = sourceSelector.getReleaseDescriptor(unirest);
-        try (FoDOnlineRemediationsFprSource source =
-                new FoDOnlineRemediationsFprSource(unirest, logger, release)) {
-            ApplyResult applyResult = RemediationsApplyHelper.apply(
-                    source, sourceCodeDirectory, logger, issueIdFilter, LOG);
-            return AviatorFoDApplyRemediationsHelper.buildOnlineResultNode(release, applyResult);
+    @Override
+    protected IRemediationsFprSource openFprSource(AviatorLoggerImpl logger, IProgressWriter progressWriter) {
```
rsenden: Same comment as for SSC

rsenden (changes requested):

rsenden (approved):

PR #1077, opened 2026-08-21
fix: `fcli ai mcp start-stdio` Add support for --server-name for custom MCP server name or default module specific name

SangameshV:
Added support for a new `--server-name` option to the `fcli ai mcp start-stdio` command. This optional parameter can only be used together with the `--module` option. When specified, it creates the MCP server for the selected module using the provided server name. If not supplied, the server name defaults to `fcli-<module-name>`.


rsenden:
@SangameshV, I guess you're still working on this, but some initial comments:
- If the MCP server is started with `--import` only (no `--module`), I think `serverName` will always be `null`; it should default to `fcli`, or maybe something like `fcli-imported-functions`
- I don't see why the `--server-name` option is restricted to the `--module` option only. With `--module`, the new default name (`fcli-<module>`) will likely usually be sufficient, whereas for `--import`, we can only generate a fixed name, so the ability to override this fixed server name makes even more sense than for `--module`
- For consistency, please check whether it makes sense to also have a `--server-name` option/configuration value for the HTTP MCP server

PR #1080, opened 2026-08-23
feat: Add FoD Session MFA Code Request Command

SangameshV:
Adds a new `fcli fod session request-mfa-code` command to request Multi-Factor Authentication (MFA) codes from Fortify on Demand (FoD) via Email or SMS delivery methods.

Triggers FoD to send a multi-factor authentication code to the specified delivery method for the given tenant and user. Use the received code with `fcli fod session login --code <code>` to complete authentication.

### DIFF
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/output/cli/mixin/FoDOutputHelperMixins.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/output/cli/mixin/FoDOutputHelperMixins.java
@@ -110,4 +110,9 @@ public class FoDOutputHelperMixins {
     public static class UploadFile extends OutputHelperMixins.TableNoQuery {
         public static final String CMD_NAME = "upload-file";
     }
+
+    @Command(aliases = "mfa")
+    public static class RequestMfaCode extends OutputHelperMixins.TableNoQuery {
+        public static final String CMD_NAME = "request-mfa-code";
+    }
 }
--- a/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionRequestMfaCodeCommand.java
+++ b/fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionRequestMfaCodeCommand.java
@@ -14,12 +14,9 @@ package com.fortify.cli.fod._common.session.cli.cmd;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.fortify.cli.common.log.LogSensitivityLevel;
-import com.fortify.cli.common.log.MaskValue;
 import com.fortify.cli.common.output.cli.cmd.AbstractOutputCommand;
 import com.fortify.cli.common.output.cli.cmd.IJsonNodeSupplier;
 import com.fortify.cli.common.output.transform.IActionCommandResultSupplier;
-import com.fortify.cli.common.session.cli.mixin.UserCredentialOptions;
 import com.fortify.cli.fod._common.output.cli.mixin.FoDOutputHelperMixins;
 import com.fortify.cli.fod._common.rest.helper.FoDProductHelper;
 import com.fortify.cli.fod._common.session.cli.mixin.FoDSessionLoginOptions;
@@ -39,10 +36,7 @@ import picocli.CommandLine.Option;
 public class FoDSessionRequestMfaCodeCommand extends AbstractOutputCommand implements IJsonNodeSupplier, IActionCommandResultSupplier {
     @Getter @Mixin private FoDOutputHelperMixins.RequestMfaCode outputHelper;
     @Mixin private FoDSessionLoginOptions.FoDUrlConfigOptions urlConfigOptions;
-    @Mixin private UserCredentialOptions userCredentials;
-    @Option(names = {"-t", "--tenant"}, required = true)
-    @MaskValue(sensitivity = LogSensitivityLevel.low, description = "FOD TENANT")
-    private String tenant;
+    @Mixin private FoDSessionLoginOptions.FoDUserCredentialOptions userCredentials;
     @Option(names = {"--delivery-mode", "-m"}, required = true)
     private FoDMfaDeliveryType deliveryMode;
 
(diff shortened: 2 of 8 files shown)


fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionRequestMfaCodeCommand.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@Command(name = FoDOutputHelperMixins.RequestMfaCode.CMD_NAME, sortOptions = false)
+public class FoDSessionRequestMfaCodeCommand extends AbstractOutputCommand implements IJsonNodeSupplier, IActionCommandResultSupplier {
+    @Getter @Mixin private FoDOutputHelperMixins.RequestMfaCode outputHelper;
+    @Mixin private FoDSessionLoginOptions.FoDUrlConfigOptions urlConfigOptions;
+    @Mixin private UserCredentialOptions userCredentials;
+    @Option(names = {"-t", "--tenant"}, required = true)
+    @MaskValue(sensitivity = LogSensitivityLevel.low, description = "FOD TENANT")
+    private String tenant;
+    @Option(names = {"--delivery-mode", "-m"}, required = true)
```
rsenden: Let's keep this option, but make it optional. If not specified, we try both.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionRequestMfaCodeCommand.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Command for requesting a Multi-Factor Authentication (MFA) code via Email or SMS.
+ * @author Sangamesh Vijaykumar
+ */
+@Command(name = FoDOutputHelperMixins.RequestMfaCode.CMD_NAME, sortOptions = false)
+public class FoDSessionRequestMfaCodeCommand extends AbstractOutputCommand implements IJsonNodeSupplier, IActionCommandResultSupplier {
+    @Getter @Mixin private FoDOutputHelperMixins.RequestMfaCode outputHelper;
+    @Mixin private FoDSessionLoginOptions.FoDUrlConfigOptions urlConfigOptions;
+    @Mixin private UserCredentialOptions userCredentials;
```
rsenden: The login command already uses an ArgGroup that declares user, password, and tenant options, right? Why not just re-use that here?

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/cli/cmd/FoDSessionRequestMfaCodeCommand.java
```diff
@@ -0,0 +1,76 @@
...
+    @Option(names = {"-t", "--tenant"}, required = true)
+    @MaskValue(sensitivity = LogSensitivityLevel.low, description = "FOD TENANT")
+    private String tenant;
+    @Option(names = {"--delivery-mode", "-m"}, required = true)
+    private FoDMfaDeliveryType deliveryMode;
+
+    @Override
+    public JsonNode getJsonNode() {
+        FoDMfaHelper.requestMfaCode(
```
rsenden: Please use a similar approach as `FoDOAuthHelper`, i.e., use `IFoUserCredentials` to pass tenant/user/password as a single method parameter.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/session/helper/FoDMfaHelper.java
```diff
@@ -0,0 +1,73 @@
...
+            requestBody.put("password", String.valueOf(password));
+            
+            unirest.post("/api/v3/multi-factor-authorization-code")
+                    .headerReplace(HttpHeader.ACCEPT, "application/json")
+                    .headerReplace(HttpHeader.CONTENT_TYPE, "application/json")
+                    .body(requestBody)
+                    .asEmpty();
+            
+            //security hardening
+            java.util.Arrays.fill(password, ' ');  // Clear original char array
```
rsenden: - Please use proper import
- This method doesn't own the `char[]`, so clearing it here is a side-effect that callers may not expect (maybe caller wants to do something else with the password). As per best practices, method side effects should be avoided. It would be better to have the caller (that declares/owns the `char[]`) decide whether the `char[]` can be cleared or is still needed for something else.
- Note that we don't explicitly clear password character arrays anywhere else, so not sure how to best make this consistent, i.e., either remove this line, or update all other commands to explicitly clear password character arrays.

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/_common/output/cli/mixin/FoDOutputHelperMixins.java
```diff
@@ -110,4 +110,9 @@ public static class GetConfigLegacy extends OutputHelperMixins.DetailsNoQuery {
     public static class UploadFile extends OutputHelperMixins.TableNoQuery {
         public static final String CMD_NAME = "upload-file";
     }
+
+    @Command(aliases = "mfa")
```
rsenden, line 114: Although it's nice to have a short alias, I'm not 100% sure whether I like this alias ('mfa' by itself has a meaning, but doesn't describe what this command does, and also maybe in the future we need to add other MFA-related commands). 

Related, can we think of a shorter full command name? I think the option on the login command is now `--code`, so maybe we can name the command `request-code`? Any other word for 'request'? Maybe ask Shajaan for suggestions?

Maybe we should reconsider this altogether, thinking first about whether we should improve the MFA handling on the `login` command. Some potential approaches:

- Restructure MFA options on `login` command: `--mfa=<code>` and `--totp=<code>`, then have separate `mfa` command to request MFA code (thus command name matching the `login` `--mfa` option name)
- Integrate MFA request into `login` command, i.e., `--mfa[=<code>]`, if no code given, we check whether there's a cached code; if not (or cached code results in denied exception due to being expired), we send the 'request MFA' request and then prompt the user to enter the MFA code

In other words, maybe we should take a step back to decide on the most user-friendly approach (while maintaining backward compatibility, possibly marking existing functionality as deprecated).

rsenden (changes requested):

PR #1085, opened 2026-09-01
Fix DAST audit eligibility and server tier handling

ankit2995:
## Summary

- Download the current SSC FPR state before DAST auditing.
- Exclude suppressed, human-audited, and Aviator-processed findings before quota submission.
- Skip stream initialization when no eligible findings remain.
- Introduce `AuditTier` for typed `GOLD`/`SILVER` handling.
- Default missing or unknown server tiers to `SILVER`.
- Harden null stream future/result handling.
- Consolidate temporary FPR download and cleanup logic.

### DIFF
--- a/fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/processor/AuditProcessor.java
+++ b/fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/processor/AuditProcessor.java
@@ -843,7 +843,7 @@ public class AuditProcessor {
 
     private TagMappingConfig.Result getDastResultConfig(AuditResponse response,
             TagMappingConfig tagMappingConfig) {
-        boolean tierOne = AuditTier.GOLD.name().equals(response.getTier());
+        boolean tierOne = AuditTier.fromServerValue(response.getTier()) == AuditTier.GOLD;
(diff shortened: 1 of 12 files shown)


fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/processor/AuditProcessor.java
```diff
@@ -842,7 +843,7 @@ private void applyDastAuditResponse(Element issueElement, AuditResponse response
 
     private TagMappingConfig.Result getDastResultConfig(AuditResponse response,
             TagMappingConfig tagMappingConfig) {
-        boolean tierOne = "GOLD".equalsIgnoreCase(response.getTier());
+        boolean tierOne = AuditTier.GOLD.name().equals(response.getTier());
```
rsenden: Why not use `AuditTier::fromServerValue` (or extra method on `AuditTier` to differentiate between case-sensitive/insensitive matching if needed)

ankit2995: Thanks @rsenden  I’ve updated both DAST string boundaries to use AuditTier.fromServerValue(...). This preserves case-insensitive matching and defaults null or unknown tiers to SILVER.

PR #1088, opened 2026-09-02
feat: Add DAST scan support to SSC ci action

jmadhur87:
Adds the ability to start ScanCentral DAST scan from the fcli SSC ci pipeline action, reusing scan settings that are already configured in SC-DAST


rsenden (changes requested):
Assuming this has been properly tested, I think this looks mostly fine. As a small improvement, I think it would be better to move the `DAST_WAIT` step down. For example, we can already download SAST logs if the DAST scan is still running.

I think even the Aviator-related calls are SAST-only for now, so we can already do SAST audit & remediations while DAST scan is still running, meaning that `DAST_WAIT` step should move after those Aviator-related steps.

Note that Aviator 26.4 will bring new DAST-related commands (and renames some SAST-related commands), so we'll likely need to make some changes once that gets released, to properly support Aviator DAST audits.

PR #1089, opened 2026-09-05
P/umadevis1/improve remediations

umadevis1:
PR updated with all the changes needed.


umadevis1:
This PR has a set of fixes that improve the remediation process.  But this does not guarantee that remediation will always work yet due to multiple nuances to overcome. As a part of fixing this, the PR also rearchitects the RemediationProcessor to separate classes to pave way for more fixes.
This PR is tested with a set of FPRs and has seen significant improvement in number of remediations done.
The unit testcases have been updated to cover all the scenarios.
**1 — Strip synthetic markers + blank lines** (`FileUtil.stripSyntheticLineMarkers`, `trimBlankLines`). fcli injects `// L37`-style markers into every line before sending code to the LLM; these get stripped from the proposed fix before it's written to `remediations.xml`, along with leading/trailing blank lines. This is the doc's stated prerequisite for near-identical detection — it converts marker-driven near-duplicates into byte-identical ones.

**P2 — Canonical hash, both sides** (`FileUtil.canonicalizeForHash`, `AuditProcessor.calculateHashBase64`, `RemediationProcessor.applyChange`). Canonical form = normalize line endings to LF + strip trailing newline, applied identically on the audit side (writing the hash) and apply side (checking it). For transition compatibility (per the doc's explicit requirement — a corrected apply side must still work against already-generated `remediations.xml`), apply tries 4 hash forms per hunk: canonical/legacy × UTF-8/FVDL-declared-source-encoding, accepts any match, and logs which form matched.

**3 — Identity dedup at hunk level, INCLUDING near-identical (whitespace/comment) matching**. Before attempting to place a hunk, fcli checks whether its target content already matches what a prior remediation in this run wrote. A remediation is `IDENTICAL` if all its hunks are already satisfied this way; `APPLIED` if any hunk is genuinely new. This is what turns "skipped, no reason" into "identical to `<instanceId>`, applied once" — the doc's single largest unexplained-skip bucket (6/13, 6/14 on the two in-scope apps).

The identity key is built from `createComparisonCode` not the raw proposed text — so the comparison is already whitespace- and comment-insensitive, per language: it strips all whitespace (`replaceAll("\\s+", "")`) and strips comments using a per-language comment-symbol lookup (`LanguageCommentMapperUtil`/`LanguagesCommentConfig` — `//`+`/* */` for C-style, `#` for script languages, `<!-- -->`/`<%-- --%>` for markup/JSP), on top of the literal-alias normalization below. This means the doc's "near-identical fixes" case (whitespace/comment differences, which it calls language-specific R&D) is already covered by the same mechanism that handles exact identity — not a separate unimplemented feature.

**4 — Patch-style application with an offset map**. This is the core structural change, replacing "always text-search":
- `Map<Path, List<AppliedChange>> appliedByFile` — per-file, per-run offset map. Each successful hunk records `AppliedChange(instanceId, originalLineFrom, originalLineTo, deltaLines)`.
- Remediations are processed widest-first (sorted by max hunk width descending), so a broad fix is recorded before narrower ones that might overlap it.
- Each new hunk's declared line range is projected through the accumulated deltas already recorded for that file this run, instead of being searched for blind.
- The projected range is then classified against the offset map: fully nested inside a prior `AppliedChange` → `SUPERSEDED` (no write needed, already covered); partially overlapping → `CONFLICTS_WITH_ANOTHER_FIX` (reported by instanceId + overlapping lines, not guessed); no overlap → verify the expected original code at the projected position and apply, or report `ANCHOR_DOES_NOT_MATCH` if it doesn't match.
- Text search is now the fallback only — used solely for files genuinely untouched this run, where the hash mismatched.
- Output model: `applied` / `identical` / `superseded` / `skipped` as four separate counters, with `skippedByReason` (Conflicts, Anchor does not match, ambiguous, source drift) explaining every remaining skip — nothing is silently dropped.

**5 — Multi-file remediation atomicity** (landed this session, on top of the above). One `<Remediation>` can bundle `<FileChanges>` for several files. Previously, if the first file's block threw (e.g. anchor mismatch), the whole remediation aborted — silently dropping a valid fix for a sibling file in the same bundle. Fixed to isolate per-file failure and let siblings still commit; a companion bug this exposed (a failed sibling's hunk getting wrongly marked "already applied," which would then cause a legitimate duplicate remediation for that same fix to be wrongly skipped) was fixed alongside it via a `PreparedFileChanges` record that returns the actual set of keys satisfied, not "all or nothing."

**Literal-alias normalization**  (`'\0'`↔`0`, `nullptr`↔`NULL`) layered on top of the comment/whitespace-stripped comparison code, so functionally-identical C c

**Boundary Token De dup Change:** in `applyChange()`, right before `NewCode`'s lines are spliced into the file, a new
`dropDuplicatedBoundaryTokens(...)` call compares the line immediately before the declared
`lineFrom` and immediately after `lineTo` (in the original file) against `NewCode`'s first/last
line. On an exact, whitespace-normalized match, that duplicate boundary line is dropped from
`NewCode` before insertion — fixing the confirmed compile-breaking duplication bug (e.g. extra
closing brace in `ProductController.java`/APM0002320). Guarded so `NewCode` can never be trimmed
to zero lines, and blank-line matches are ignored (a blank line isn't a boundary token). 

**Exact Line disambiguation**  Changed`fuzzySearchContext()` and `fuzzySearchOriginalCode()` now take the FPR's declared
`LineFrom`/`LineTo`, projected through this run's offset ledger (the `shift` calculation
previously computed only inside the offset-projection block is now computed unconditionally in
`applyChange()` and passed down). When either method's underlying fuzzy search finds more than one
candidate, it no longer throws `SOURCE_CONTEXT_AMBIGUOUS`/`ORIGINAL_CODE_AMBIGUOUS` immediately —
it first checks whether exactly one candidate's position exactly equals that projected/declared
position. If so, that candidate is used. If zero or more than one candidate coincide with it,
behavior is unchanged: still throws ambiguous, same as before. Exact-equality check against
data already trusted elsewhere in this class (the same ledger that powers hash-match/offset-
projection) consistent with the project's no-heuristic rule.

kireetivar (approved):

umadevis1:
Reverted a partial remediation fix.  If a remediation has multiple hunks, we either apply all or none.

Also capturing here the remediation states


# Remediation classification states 

Each remediation processed by `RemediationProcessor.classifyAndApply` ends up in exactly one of
these states. They are checked in this order for every remediation, and remediations are
processed **widest-hunk-first** — so a narrow fix can only classify as Superseded /
Possibly-remediated / Conflicts if a broader one already landed earlier in the same run.

| State | Field | When it's assigned | Example |
|---|---|---|---|
| **Identical** | `identicalRemediations` | Every hunk in the remediation has the same file, line range, and normalized NewCode as a hunk already applied earlier in this run. Nothing is written — just counted. | Remediation A applies NewCode `"REPLACED"` at lines 2-2. Remediation B proposes the same fix at the same lines, worded `"REPLACED // note"` — normalizes to the same content, so B is "fully identical to prior remediation(s) [A]". |
| **Superseded** | `supersededRemediations` | The remediation's range is fully nested inside an already-applied change, and that prior change's written content already contains what this remediation proposes. Nothing is written. | A broad fix rewrites lines 1-3 to `"before\nREPLACED\nafter"`. A separate, narrower remediation at line 2 proposing `"REPLACED"` is superseded — the broad fix already did it. |
| **Possibly-remediated** | `possiblyRemediatedRemediations` | Same nesting as Superseded, but the prior fix's content does *not* match what this remediation proposes — some other fix already touched that spot, just not provably the same one. Nothing is written. | Broad fix rewrites lines 1-3 to `"wideline1\nwideline2\nwideline3"`. A narrower remediation at line 2 proposing `"NARROW_DIFFERENT"` doesn't match → possibly-remediated, not applied. |
| **Conflicts** | rolled into `skippedRemediations`, under reason "Conflicts with another fix" | The remediation's range partially overlaps an already-applied change — neither range contains the other, so coverage is genuinely ambiguous. Skipped rather than guessed. | Fix A rewrites lines 2-3. Fix B (processed after) rewrites lines 3-4 — line 3 overlaps but neither fully contains the other → B conflicts, skipped. |
| **Applied** | `appliedRemediations` | None of the above apply — the fix is actually located and written to disk. | Normal case. |
| **Skipped** (other reasons) | `skippedRemediations` | Location/write failed for a reason unrelated to prior fixes (context/original code not found, file missing, etc). | Context text no longer exists anywhere in the file. |

umadevis1:
This PR also addresses an audit problem where source files were not found due to the following scenario


**Scenario**:
If an FPR is a pure WebInspect/DAST scan, it has no src-archive worth indexing, so return an empty map rather than trying (and failing) to parse a nonexistent index.xml.
The bug is that it only checked "does webinspect.xml exist" — it never checked whether the FPR also had a real SAST scan (audit.fvdl + src-archive). SSC supports merged Application Versions where a SAST scan and a DAST scan are uploaded into the same AV and get combined into one FPR. In that merged case, the FPR legitimately contains both webinspect.xml and a fully populated audit.fvdl/src-archive. Before the fix the code had the webinspect.xml, assumed "this is DAST, no source," and threw away the entire source map — even though a valid src-archive was sitting right there.
Because sourceFileMap is memoized (if (sourceFileMap == null)) and first populated inside validate() at the very start of the audit flow, that wrong empty map got locked in for the rest of the FPR's lifetime. Every later lookup (FileUtils.resolveFullPath → getLineFromFile/getFragmentFromFile, used by every trace node in every issue) silently missed and returned "". Reason why Aviator's comments kept saying "no source code was provided" for the SAST issues, while validate() itself never threw an error — it has the correct AND logic (!audit.fvdl exists && webinspect.xml exists → reject), so it proceeded, masking the problem downstream.
**Does it occur when webinspect.xml is absent?**
No. If there's no webinspect.xml at all, the ternary always takes the loadSourceFileMap() branch regardless of the old or new code — the two versions are only different when webinspect.xml is present. So:
Pure SAST FPR (no webinspect.xml): unaffected before and after — always worked correctly.
Pure DAST FPR (webinspect.xml, no audit.fvdl): unaffected by the fix — still correctly returns an empty map (and validate() rejects it anyway with the "DAST scan" error).
Merged SAST+DAST FPR (webinspect.xml and audit.fvdl both present): this was the broken case — now fixed, since dastOnly evaluates false and loadSourceFileMap() runs.

ankit2995 (approved):

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -156,7 +159,8 @@ JsonNode processAllAviatorArtifacts(OffsetDateTime sinceDate) {
             String action = appliedRemediations > 0 ? "Remediation-Applied" : "No-Remediation-Applied";
             return AviatorSSCApplyRemediationsHelper.buildAggregatedResultNode(
                     appVersionId, artifactsProcessed, artifactsSkipped,
-                    totalRemediations, appliedRemediations, skippedRemediations, allModifiedFiles, skippedByReason, action);
+                    totalRemediations, appliedRemediations, identicalRemediations, supersededRemediations,
```
rsenden: Although previous code already passed a large number of arguments to this function, even more args have been added now, many with the same type, which makes it very easy to accidentally provide arguments in the wrong order. Please look into options for improving this (both here and other code in this PR). 

Some potential ideas:
- Introduce a new `AggregatedResultNode` class with Lombok `@Builder` annotation and builder utility methods for increasing existing counts; at the start of this method, you'd create the builder, then have the processing logic for example call `resultBuilder.addTotalRemediations(metric.totalRemediations())`, or even better, a single `resultBuilder.add(metric)` method call; once processing is complete, you'd call the `build()` method on the builder, and then call some method on the result to produce final output.
- I guess `metric` may already use builder pattern, so maybe it can be reused for both single-result metrics and aggregated result metrics. Very similar to the above, you'd instantiate MetricBuilder, call `aggregatedMetricBuilder.add(metric)`, and finally call `AviatorSSCApplyRemediationsHelper.buildAggregatedResultNode(..., aggregatedMetricBuilder.build())`

Both ideas above also allow for removing all the individual counter variables from this method, as these are managed by the builders.

fcli-core/fcli-aviator/src/main/java/com/fortify/cli/aviator/ssc/cli/cmd/AviatorSSCApplyRemediationsCommand.java
```diff
@@ -185,12 +189,14 @@ JsonNode processFprRemediations(SSCArtifactDescriptor ad) {
...
                 try (FprHandle fprHandle = new FprHandle(fprPath)) {
                         var remediationMetric = ApplyAutoRemediationOnSource.applyRemediations(fprHandle, sourceCodeDirectory,
-                            sourceEncodingsMixin.getSourceDecoder(), logger);
+                            sourceEncodingsMixin.getSourceDecoder());
                     String status = remediationMetric.appliedRemediations() > 0
                         ? "Remediation-Applied"
                         : "No-Remediation-Applied";
                     return AviatorSSCApplyRemediationsHelper.buildResultNode(ad, remediationMetric.totalRemediations(),
-                            remediationMetric.appliedRemediations(), remediationMetric.skippedRemediations(),
+                            remediationMetric.appliedRemediations(), remediationMetric.identicalRemediations(),
```
rsenden: Same comment as elsewhere regarding large number of method arguments

fcli-core/fcli-fod/src/main/java/com/fortify/cli/fod/aviator/cmd/FoDAviatorApplyRemediationsCommand.java
```diff
@@ -85,13 +85,15 @@ private JsonNode processFprRemediations(UnirestInstance unirest, FoDReleaseDescr
...
                 var remediationMetric = ApplyAutoRemediationOnSource.applyRemediations(fprHandle, sourceCodeDirectory,
-                    sourceEncodingsMixin.getSourceDecoder(), logger);
+                    sourceEncodingsMixin.getSourceDecoder());
                 LOG.info("Applied remediation {}", remediationMetric.appliedRemediations());
                 LOG.info("Total remediation {}", remediationMetric.totalRemediations());
                 String status = remediationMetric.appliedRemediations() > 0 ? "Remediation-Applied" : "No-Remediation-Applied";
                 return AviatorFoDApplyRemediationsHelper.buildResultNode(rd, remediationMetric.totalRemediations(),
-                    remediationMetric.appliedRemediations(), remediationMetric.skippedRemediations(), remediationMetric.modifiedFiles(),
-                    remediationMetric.skippedByReason(), status);
+                    remediationMetric.appliedRemediations(), remediationMetric.identicalRemediations(),
```
rsenden: Same comment about number of method args

rsenden (changes requested):
I think Frans already partially reviewed the code and functionality; so I only reviewed the changes in the `fcli-aviator` module (not `fcli-aviator-common`). Some of my comments may also apply to methods in `fcli-aviator-common` code, so please pdate there as well if applicable.

fcli-core/fcli-aviator-common/src/main/java/com/fortify/cli/aviator/fpr/remediation/model/RemediationMetric.java
```diff
@@ -0,0 +1,85 @@
...
+ * without notice.
+ */
+package com.fortify.cli.aviator.fpr.remediation.model;
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+public final class RemediationMetric {
```
rsenden, line 27: Why didn't you use Lombok `@Builder` annotation? That would save a lot of boilerplate code, and still allows for builder customization like adding the `add` method.

rsenden (approved):

PR #1091, opened 2026-09-07
fix: convert FoD identifier fields from int to long to prevent overflow

jmadhur87:
Fixes potential overflow bugs in FoD identifier handling. Several FoD IDs (report, release, user, group, attribute, scan, entitlement) were stored as Integer/int, which breaks once an ID grows past ~2.1 billion. Converted them to long. 

Affects several commands: release update/create (--owner/--app-owner), app create (--owner), issue update (--user), group/user update (--add-users/--remove-users/--add-groups/--remove-groups), sast/mast/dast-scan setup/start (--entitlement-id), attribute and report get/update/delete (--attribute-id/--report-id), plus output field types on several list/get commands.


rsenden (changes requested):
I think this looks mostly fine, just wondering though why you changed `FoDAppDescriptor::applicationId` from `String` to `Long`, whereas you didn't do the same for `FoDReleaseDescriptor::applicatonId`, nor for the `releaseId` and `microserviceId` fields in the various descriptors.

Please ensure we're using a consistent approach for these id's; either treat them all as strings, or treat them all as longs.

jmadhur87:
@rsenden For consistency, reverted the applicationId from long to string.

